PR #2, opened 2021-09-02
Update UI to be responsive

makermelissa:


tannewt (approved):
Awesome! Thank you for making this much better. Will test it and follow up.

PR #5, opened 2021-10-20
Add File Operation Functions

makermelissa:
This adds the New, Open, Save As buttons in along with expected functionality. Currently Save + Run only saves.


makermelissa:
Oh yeah, https://github.com/adafruit/ble-file-transfer-js/pull/3 will need to be merged before this passes CI.

makermelissa:
I nudged it, but it doesn't seem to realize that the package exists. Is there something else I'm supposed to do like a release?

Issue #8, opened 2021-11-17
Warn users about experimental chrome flags 


beaufortfrancois:
https://learn.adafruit.com/wirelessly-code-your-bluetooth-device-with-circuitpython/device-setup and https://code.circuitpython.org/ recommends users to enable two experimental flags in Chrome.

The `enable-experimental-web-platform-features` is a flag that affects all websites in Chrome. It should not be enabled unless user knows their browsing experience in Chrome will be impacted.

The `enable-web-bluetooth-new-permissions-backend` flag on the other hand only affects Bluetooth pickers in Chrome. That said, this may break other websites that use Web Bluetooth as well.

Are those flags **really** required?
If not, please remove them from those pages.

beaufortfrancois:
If `enable-experimental-web-platform-features` is to enable the Web Bluetooth API on Linux, you may want to detect this in JavaScript so that non-Linux users don't see these potentially harmful instructions.

```js
if (
  !navigator.bluetooth &&
  navigator.userAgentData.platform == "Linux" &&
  navigator.userAgentData.brands.some((item) => {
    ["Chromium", "Google Chrome"].includes(item.brand);
  })
) {
  // Ask user to enable the experimental web platform features flag and
  // warn them it's going to impact their browsing experience on all websites.
}
```

Regarding the `enable-web-bluetooth-new-permissions-backend`, is it because you use the new `navigator.bluetooth.getDevices()` that has not shipped yet? See https://www.chromestatus.com/feature/4797798639730688

beaufortfrancois:
@tannewt @makermelissa

makermelissa:
I'm planning on testing this out later today. It's very possible that `enable-experimental-web-platform-features` may only be required on certain OS's or even not at all and I'll adjust things accordingly.

> Regarding the `enable-web-bluetooth-new-permissions-backend`, is it because you use the new `navigator.bluetooth.getDevices()` that has not shipped yet? See https://www.chromestatus.com/feature/4797798639730688

That sounds right. I believe the `enable-web-bluetooth-new-permissions-backend` was for re-establishing the bond.

makermelissa:
So far `enable-experimental-web-platform-features` is required on linux to work.

makermelissa:
[...]

PR #9, opened 2021-11-19
Update prose around flags

beaufortfrancois:
Following https://github.com/circuitpython/web-editor/issues/8 discussion, I believe we should make it clear what flags are for.

@makermelissa Please review


makermelissa (approved):
Thanks for making these changes. It's also good to know it works on Edge browsers.

PR #14, opened 2022-02-03
Update Web Bluetooth flag name

beaufortfrancois:
I've just added a flag to enable the Web Bluetooth API on Linux at `chrome://flags/#enable-web-bluetooth`

See https://twitter.com/quicksave2k/status/1489150418319990785

@makermelissa

### DIFF
--- a/index.html
+++ b/index.html
@@ -67,7 +67,7 @@
                                         href="about://flags/#enable-web-bluetooth-new-permissions-backend">about://flags/#enable-web-bluetooth-new-permissions-backend</a> flag.
                                 </p>
                                 <p>
-                                    On Linux only, turn on Web Bluetooth by enabling with the experimental <a
+                                    On Linux only, enable the experimental <a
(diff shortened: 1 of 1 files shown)


makermelissa:
Hi, on my system running Chrome 98.0.4758.80, this new option does not appear. My system is a bit older and this leads me to believe, that only the absolute newest systems would show this option. Perhaps you could reword it so that a user checks the new flag first and if they can't find it, they fall back to the existing flag for systems that may be a little older.

makermelissa (changes requested):
Hi, on my system running Chrome 98.0.4758.80, this new option does not appear. My system is a bit older and this leads me to believe, that only the absolute newest systems would show this option. Perhaps you could reword it so that a user checks the new flag first and if they can't find it, they fall back to the existing flag for systems that may be a little older.

beaufortfrancois:
@makermelissa I've edited prose. Let me know what you think.

makermelissa (approved):
Sounds good. Thanks.

PR #16, opened 2022-07-28
Refactored to accommodate additional workflows

makermelissa:
This also fixes a couple of minor bugs I found while reworking the code like trying to "save as" with a filename that doesn't exist after clicking one that does.


tannewt (approved):
Looks good! I'm excited for this!

PR #17, opened 2022-07-29
Update index.html

sctv:
when I click on the about://flags/#enable-web-bluetooth-new-permissions-backend link on https://code.circuitpython.org/ on windows10 chrome, it loads about:blank#blocked
chrome://flags/#enable-web-bluetooth-new-permissions-backend might be a better link.

### DIFF
--- a/index.html
+++ b/index.html
@@ -64,7 +64,7 @@
                                 </p>
                                 <p>
                                     To get the best experience, enable the experimental <a
-                                        href="about://flags/#enable-web-bluetooth-new-permissions-backend">about://flags/#enable-web-bluetooth-new-permissions-backend</a> flag.
+                                        href="chrome://flags/#enable-web-bluetooth-new-permissions-backend">chrome://flags/#enable-web-bluetooth-new-permissions-backend</a> flag.
                                 </p>
                                 <p>
                                     On Linux only, enable the experimental <a


makermelissa:
This was changed in #9 so it would work on Microsoft Edge Browsers too. Copying/Pasting into the URL works with the way it is now. If we detected chrome vs edge and displayed the appropriate url, I'd be ok with that.

makermelissa (changes requested):
Please see previous comment. This change breaks compatibility with Microsoft Edge Browsers.

sctv:
Thanks for the information, will consider it closed as I don't [currently] have the knowledge to code to detect which browser the page is being displayed on.

A work-around/stopgap might be to suggest that the user cuts and pastes the URL in chrome, as it might be mysterious otherwise.

makermelissa:
Ok, well I'm currently working on a huge update to implement web workflow and if it seems easy enough, I might just make that part of the update.

makermelissa:
I just tested by updating with chrome:// links and they don't do anything when clicked and if you open them in a new tab, it has the same blocked issue.

sctv:
Thanks... there is probably more to linking to settings from pages in chrome. 
Cut and paste seems to be the way to do it currently. 
Sorry for taking up your time.

makermelissa:
No worries. :)

PR #19, opened 2022-08-12
Added Web Workflow

makermelissa:
This is a huge update to add Web Workflow and revises a lot of the structure of the Code Editor. To use web workflow, you will need to do the following:

1. Make sure the CIRCUITPY drive is disabled. See [this guide](https://learn.adafruit.com/customizing-usb-devices-in-circuitpython/circuitpy-midi-serial#circuitpy-mass-storage-device-3096583). If you don't do this, the device will be in read-only mode.
2. Hard reset the device after disabling the CIRCUITPY drive.
3. Make sure you are connected with **http** and not **https**. The main reason is that Web Workflow on the CircuitPython device does not support https nor certificates. 
4. Append #backend=web&host=_[hostname]_ with [hostname] being something like **cpy-XXXXXX.local**

It should connect automatically. BLE functionality is a little different in that the connect button brings up a dialog now.


makermelissa:
I'm going to merge this in because pretty much nobody is regularly using the site, Scott is off today, and I already had verbal approval, so it will allow people to do some testing. A further update will be created, so it will wrap up any fixes needed from this PR.

PR #33, opened 2022-10-19
Added `hover` effect

Franqsanz:
Added the `hover` effect and a scss variable.


assets/sass/base/_base.scss
```diff
@@ -70,6 +70,10 @@ h5 {
     }
```
makermelissa, line 70:
```suggestion
    &:hover {
        background-color: darken($purple, 15%);
    }
}
```

makermelissa (changes requested):
You should be able to just add a hover to the button class and use the sass darken function instead of creating another variable.

assets/sass/base/_base.scss
Franqsanz: You're right! I fix it now.

makermelissa (approved):
Bracing got a bit messed up, but I fixed it. Thank you for this.

makermelissa:
No worries. That's how we learn. It's great to have other contributing.

Franqsanz:
Thank you very much!

Issue #15, opened 2022-06-22
Edit CIRCUITPY files for USB workflow


tannewt:
Use the directory picker to get access to a CIRCUITPY drive: https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/webkitdirectory

makermelissa:
Unfortunately I don't think this will work the way you may be thinking it will.

While it sort of works for selecting a folder with a couple of caveats (mostly the scary warning that it will upload a huge number of files from the path that is selected), the folder location isn't really useful as you can't specify where you would like to save files on a user's system. I mean you might be able to do a scan of the files on the device to get an idea of the layout, but since it is subject to change after that, this wouldn't be very useful.

For now, I think an approach where we open a single file and download a single file will work best unless we think of something that's a bit more clever that doesn't interfere with web security.

If I'm incorrect and there is a fancy way to do this, feel free to provide an example of another site that is able to work in a similar fashion. Closing for now.

makermelissa:
Reopening because this looks like using https://developer.mozilla.org/en-US/docs/Web/API/FileSystemDirectoryHandle might be a better approach, but basically the same idea.

PR #34, opened 2022-10-30
Reference error and error in removing .py from the path 

runningdeveloper:
Hi I was trying out the USB workflow, with  Save + Run. I got a ref error in the console trying to get changeMode from the script.js:
image
So I've added an export for changeMode and MODE_SERIAL (not sure if its good, but snowpack seems to sort it out).
Then when trying to run a file "test.py", it would try to run "import es" instead of "import test". So I think the extension substring is a typo. The following seemed to work to remove .py:
```
path.slice(0, -3)
```
 
Hope this helps a little bit. It's going to be nice to have a quick way to test code from just the browser.

FYI: I don't 100% understand the Save + Run for USB. It doesn't save the files to the device. I suppose it's related to #15, I wonder if we could also add a RUN only command that will reset, then paste the code in the editor to the REPL to run tests quickly.

### DIFF
--- a/assets/js/constants.js
+++ b/assets/js/constants.js
@@ -1,0 +1,3 @@
+// place to put global constants
+export const MODE_EDITOR = 1;
+export const MODE_SERIAL = 2;
--- a/assets/js/script.js
+++ b/assets/js/script.js
@@ -9,6 +9,7 @@ import {USBWorkflow} from './workflows/usb.js';
 import {CONNTYPE, CHAR_CTRL_D, isValidBackend, getBackendWorkflow, getWorkflowBackendName} from './workflows/workflow.js';
 import {ButtonValueDialog, MessageModal} from './common/dialogs.js';
 import {sleep, isLocal, switchUrl, getUrlParam} from './common/utilities.js';
+import {MODE_EDITOR, MODE_SERIAL} from './constants.js';
 
 var terminal;
 var fitter;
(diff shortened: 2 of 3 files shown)


makermelissa (changes requested):
Thanks for finding this. As for the import, I'd rather not have anything imported from `script.js` as this is the outermost script. The way I've been doing it is in the loadWorkflow function, any functions like changeMode are usually [passed in](https://github.com/circuitpython/web-editor/blob/main/assets/js/script.js#L259-L267) and [assigned](https://github.com/circuitpython/web-editor/blob/main/assets/js/workflows/workflow.js#L66-L72). It looks like I missed this one.

As for the path, I think it was supposed to take an absolute path like `/myproject/test.py` and convert that to `import myproject.test`. Perhaps it changed since it was working like that. In either case, we should probably check for a leading slash for absolute paths and also allow for relative paths like in your case.

> FYI: I don't 100% understand the Save + Run for USB. It doesn't save the files to the device. I suppose it's related to #15, I wonder if we could also add a RUN only command that will reset, then paste the code in the editor to the REPL to run tests quickly.

Yeah, it's a bit rough at the moment, but hopefully we can get it working smoother by fixing #15.

runningdeveloper:
Hi thanks for the feedback. I've fixed the function to be rather passed in. I also needed the MODE_SERIAL constant. I thought maybe putting it in some kind of global file would be good or should I pass it in the workflow.init too?

I'm ok if you would rather remove this PR and make the changes, I learnt something and got a taste of the project so all good.

makermelissa (approved):
Ok yeah. Having a constants file is perfect. Thanks.

PR #39, opened 2022-11-21
Fix non-standard port usage and Run Code import bugs

### LINKED ISSUE #37: Use web workflow port to connect to the board
The web workflow doesn't take into account the board's port, and always tries to connect to 80.
When the board is configured uses a different port, it can't connect back to it. (`CIRCUITPY_WEB_API_PORT`)

> Unable to connect. The device 192.168.1.92 was not found. Be sure it is plugged in and set up properly.

(from https://github.com/adafruit/circuitpython/issues/7168 )

### LINKED ISSUE #38: `import .code` appearing in REPL after hitting `Save & Run` on web device
After hitting Save & Run on a device, I'm seeing:

image

makermelissa:
Fixes #37. Fixes #38.


tannewt (approved):
Thank you!

PR #42, opened 2022-12-03
USB Workflow Initial Rewrite

### LINKED ISSUE #15: Edit CIRCUITPY files for USB workflow

### LINKED ISSUE #21: USB Workflow: Disconnecting in cause the page to freeze.
Test condition:
- Windows 10 with Chrome 106.0.5249.119 
- Seeed Xiao RP2040
- CircuitPython 7.3.3

Procedure:
- Connect the microcontroller with USB workflow
    - without opening any file
    - from browser console, seeing the serial output
        - which confirms that the code is running
- Click on the `Disconnect` button on the top-right corner
- Result: the browser tab freezes.

Tried:
- checked browser console, seeing log that the same info is repeated fast.
    - probably there is a dead loop caused by disconnection.

image

makermelissa:
This is a bit rough around the edges, but functional.

Fixes #15. Fixes #21.

Oh yeah, it also fixes an issue with xterm not resizing properly sometimes.


tannewt (approved):
Thank you! I'm excited to try it!

PR #46, opened 2022-12-22
Run CircuitPython code via REPL and get back result

makermelissa:
No rush on this one. You can review after you get back from the holidays. 

This adds the ability to run code via the REPL and get back the results, which need more testing at this point because I think it might have issues when the. board sits for a while before being connected. This also adds the ability to handle partial connections better. If time allows before this is reviewed, I'll add some more UI related fixed such as checkmarks to the dialog.

### DIFF
--- a/assets/js/common/usb-file-transfer.js
+++ b/assets/js/common/usb-file-transfer.js
@@ -1,3 +1,5 @@
+import { get, set } from 'https://unpkg.com/idb-keyval@6.2.0/dist/index.js';
+
 class FileTransferClient {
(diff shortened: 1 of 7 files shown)


tannewt (approved):
Looks good! I'm excited to try it. Will let you merge in case you have anything else to do beforehand.

tannewt:
@makermelissa want to merge this?

makermelissa:
I was holding off because I was focusing on the CP Installer, but we can merge anyways.

Issue #47, opened 2023-01-14
Blinka spinner over serial


tannewt:
The blinka spinner currently stays over the serial pane.

![Peek 2023-01-13 15-33](https://user-images.githubusercontent.com/52649/212440027-15e45943-ee9f-4ca4-af9d-313a25674957.gif)

makermelissa:
Fixed via #49.

PR #49, opened 2023-01-24
Run CircuitPython code via REPL and get back result (attempt 2)

makermelissa:
This is just a duplicate of #46 so far. I want to test out web workflow and see what broke and will add on to this.


makermelissa:
Ok, making progress on this. The code that interacts with the REPL has been refactored into a separate file now. I'm still tracing down the bug that's causing Web Workflow to not run.

makermelissa:
Ok, finally fixed the bug. I just want to do some additional testing before changing it to a regular PR. I may also publish the REPL component in its own repo first.

makermelissa:
I may throw a quick fix in for #47 as well.

makermelissa:
Found and fixed a couple more bugs. It looks like this fixes #47 without needing to add any additional code.

tannewt:
Let me know when its ready for review! Thanks for the fixes.

makermelissa:
It's super close. I'm just cleaning up the flow a bit. Right now it attempts to load the FileSystemAccess API handle when you click the button and doesn't give you a chance to select a different folder unless it can't connect, which could be problematic. After that's solved, it should be ready.

makermelissa:
Ok, it seems to be working well now.

makermelissa:
A lot of functionality was moved to https://github.com/adafruit/circuitpython-repl-js so it could be used by the CircuitPython Installer.

tannewt (approved):
Looks good to me! I'll let you merge when you are ready.

makermelissa:
Sounds good. It should be good to go.

PR #69, opened 2023-03-29
fix typo

Kriechi:
Fixing a typo.


makermelissa (approved):
Thanks for fixing. :)

PR #70, opened 2023-03-29
update tag line to mention all currently available connectivity options

Kriechi:
Code Editor is no longer **only** supporting "nrf52-based devices".
This PR updates the tag line / description to a more generic mention of different connectivity options.


makermelissa (approved):
Thanks. Nice catch. It looks like GitHub Actions is currently down, so I'll need to wait for that before merging these PRs.

PR #71, opened 2023-03-29
hide all other steps if no web bluetooth support is available

Kriechi:
The previous behaviour was confusing to users if the steps and buttons are visible but greyed out.

Also fixing a typo.

same as #68 but for bluetooth.


makermelissa (approved):
Looks good. Thanks.

PR #72, opened 2023-03-30
use same layout for editor-bar and serial-bar

Kriechi:
Previously, one used flexbox and the other not, causing a small pixel offset when switching modes.

### DIFF
--- a/assets/sass/layout/_header.scss
+++ b/assets/sass/layout/_header.scss
@@ -129,7 +129,7 @@
     }
 }
 
-#editor-bar {
+#editor-bar, #serial-bar {
     display: flex;
 }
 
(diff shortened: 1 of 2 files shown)


makermelissa:
Have you verified the serial terminal is working properly? It was a bit unreliable for a moment, which is why the inconsistency was there. I believe this was patched in the version it's currently pointing to though.

makermelissa (changes requested):
I'd like to verify this isn't going to break the serial stuff

Kriechi:
Yes, I have been using the terminal all day.
This PR should only change the two menu bars on top by using the same div-nesting and styling between editor and terminal mode.

makermelissa (approved):
Ok, works for me. :)

PR #73, opened 2023-03-30
address all buttons with classes instead of mix out of class+id

Kriechi:
address all buttons with classes instead of mix out of class+id

No behaviour or UI change - just a bit of code hygiene to have the same conventions across all buttons.

### DIFF
--- a/assets/js/script.js
+++ b/assets/js/script.js
@@ -22,16 +22,16 @@ workflows[CONNTYPE.Ble] = new BLEWorkflow();
 workflows[CONNTYPE.Web] = new WebWorkflow();
 workflows[CONNTYPE.Usb] = new USBWorkflow();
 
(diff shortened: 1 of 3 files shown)


makermelissa:
Have you verified the JS code still works? The reason it's mixed is because the ones that were classes were in 2 places (the mobile menu and the regular menu bar) and the JS code had to be updated for those to work properly.

makermelissa (changes requested):
I'd like to verify that you tested all the changed menu items.

Kriechi:
Yes, I did. There are four buttons that changed:
* the two mode buttons for Editor and Serial
* the Info button
* the Restart button on the Serial pane

I also grepped for all occurrences in the source code - pretty simple on 2 or 3 references for each.

makermelissa (approved):
Awesome. Thank you.

PR #74, opened 2023-03-30
fix incomplete serial data due to REPL partial parsing and mutating of data field

Kriechi:
caused by https://github.com/adafruit/circuitpython-repl-js/blob/main/repl.js#L108

Not sure if this should be addressed in this library directly or worked around here instead.
I tried updating web-editor to use the latest v1.2.1 of repl.js, but this just caused different problems and new/unexpected behaviour for me.

For now, this workaround seems like the most reasonable fix, also given that it also works for CircuitPython 7.x, while the repl.js at least states in the readme it only supports v8 at this time.

### DIFF
--- a/assets/js/workflows/workflow.js
+++ b/assets/js/workflows/workflow.js
@@ -114,9 +114,10 @@ class Workflow {
     }
 
     async onSerialReceive(e) {
-        await this.repl.onSerialReceive(e);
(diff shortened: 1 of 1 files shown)


makermelissa (approved):
Thanks, this works for now. :)

PR #76, opened 2023-03-30
debug log: cleanup lines and fix broken indendation

Kriechi:
The current behaviour of debug log messages caused some not so pretty output.

If the current serial output is a partial line, i.e., a `>>>` prompt without newline at the end, the debug line would be in the same line. This PR now always puts it on a fresh line by itself.

Previously, there was a large horizontal indentation after the log line, now the new line starts at column 0 as expected.

**Before:**
```
>>> Foo**connected**
         <some space here>
```

**After:**
```
>>> Foo
**connected**
<curser column 0>
```

### DIFF
--- a/assets/js/script.js
+++ b/assets/js/script.js
@@ -318,7 +318,8 @@ async function changeMode(mode) {
 }
 
 async function debugLog(msg) {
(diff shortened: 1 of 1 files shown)


makermelissa (approved):
Looks good. Thanks.

PR #78, opened 2023-03-30
use loader animation while getting device UID, which might hang for a few seconds before timeout

Kriechi:
The repl.js library has a timeout https://github.com/adafruit/circuitpython-repl-js/blob/main/repl.js#L36 which aborts and repl.runCode call. While this is happening, there is no UI feedback to the user that it is actually doing something.

This PR simply shows the loader animation while a runCode is active.

Not sure what the expected behaviour of getting the device ID is, or what it is used for, but for my test devices and CircuitPython app, this failed frequently and caused a lot of waiting for me until it eventually timed out and I could move on with clicking through the connecting dialog.

### DIFF
--- a/assets/js/workflows/usb.js
+++ b/assets/js/workflows/usb.js
@@ -239,7 +239,7 @@ class USBWorkflow extends Workflow {
             await this.writer.ready;
         }
 
-        await this._getDeviceUid();
+        await this.showBusy(this._getDeviceUid());
 
         this.updateConnected(CONNSTATE.partial);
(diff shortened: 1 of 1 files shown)


makermelissa:
Can you also try updating the version of the repl.js and testing. I did some updates to the repl lib to make it work better while working on the circuitpython installer, but haven't had a chance to test on the editor.

makermelissa (approved):
Ok, it looks like you did do this already.

PR #79, opened 2023-03-30
auto-render web links in terminal as clickable hyperlinks

Kriechi:
This PR makes everything that looks like a URL in the terminal output to be clickable in the browser which then opens the link in a new tab. No visual change except when you mouse-hover over the URL itself.


makermelissa (approved):
I fixed a merge conflict, but otherwise it looks good.

PR #80, opened 2023-03-30
terminal auto-resizing and fitting

Kriechi:
Previously, the first time one changes from editor to terminal mode, the terminal itself did not resize as the DOM has not updated yet. If one then resizes the whole window, it will trigger correctly. Also if you cycle back to editor-terminal-editor-terminal it works.

This PR fixes this by waiting for the DOM to update and render correctly.


makermelissa (approved):
Looks good to me. Previous versions of the fit add-on required an even more hacky approach.

PR #81, opened 2023-03-30
Button layout and folder logging

Kriechi:
Previously, the only visible button in *Step 2* was shifted to the right due to a margin on the button as it was not the first element in the DOM, although it was the first and only visible element.

This PR addresses this by adding classes to apply the same styling / margin-left based on the state of the UI. Once a folder is selected, the primary button is purple, while the other button is using the gray color to indicate that it is a secondary actions. This should guide users better through the step-by-step process of first clicking to connect serial, then clicking to open the folder selector dialog, and then clicking to confirm the selected folder.

This PR also updates the log messages to be more readable.

Both buttons are now vertically left aligned:
Screenshot 2023-03-30 at 19 45 17

New gray color scheme for secondary button after a folder has been selected:
Screenshot 2023-03-30 at 19 44 46


assets/js/workflows/usb.js
```diff
@@ -246,11 +248,12 @@ class USBWorkflow extends Workflow {
         // At this point we should see if we should init the file client and check if have a saved dir handle
         this.initFileClient(new FileTransferClient(this.connectionStatus.bind(this), this._uid));
         const fileClient = this.fileHelper.getFileClient();
-        console.log("Got file client");
         const result = await fileClient.loadSavedDirHandle();
-        console.log(result);
         if (result) {
+            console.log("Successfully loaded directory:", fileClient.getWorkingDirectoryName());
             this._hostFolderChanged();
```
makermelissa, line 254: If you're making _hostFolderChanged() asynchronous, please add an await before `this._hostFolderChanged();`

assets/js/workflows/usb.js
```diff
@@ -208,15 +208,17 @@ class USBWorkflow extends Workflow {
         this.onConnected();
     }
 
-    _hostFolderChanged() {
+    async _hostFolderChanged() {
```
makermelissa, line 211: If you're making this asynchronous, please add an await before each call to this like on line #202.

makermelissa (changes requested):
Please add awaits before calls to _hostFolderChanged() if changing to async unless you have a good reason not to. Otherwise, it can lead to race conditions and related bugs.

assets/js/workflows/usb.js
Kriechi: Done.

assets/js/workflows/usb.js
Kriechi: Done.

makermelissa (approved):
Thank you. Looks good.

PR #82, opened 2023-03-30
serial mode: add Clear button to wipe current terminal content

Kriechi:
Screenshot 2023-03-30 at 22 56 32


makermelissa (approved):
Good idea. Thanks.

PR #87, opened 2023-04-23
fix saveAs filename handling

Kriechi:
The Save-As function currently does not update the file name in the UI and causes confused state of the app.
This PR fixes this problem.

**screen recording of the bug:**

https://user-images.githubusercontent.com/114300/233833019-dc06b1b6-9aa3-480f-90f2-b640ec89484b.mp4

**screen recording with the fix:**

https://user-images.githubusercontent.com/114300/233833027-5d458213-c4bf-40c4-8edc-631ea1218be4.mp4

### DIFF
--- a/assets/js/workflows/workflow.js
+++ b/assets/js/workflows/workflow.js
@@ -184,22 +184,22 @@ class Workflow {
 
         if (!path) {
             console.log("File has not been saved");
-            return;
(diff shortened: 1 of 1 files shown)


makermelissa (approved):
Thanks, looks good.

PR #88, opened 2023-04-23
Add save button

Kriechi:
requires #87 to work properly. You can review only the most recent commit in this PR.

Editing config files or other non-code files was previously a challenge because the only options to save where "Save As" which required a few more (unnecessary) clicks to complete, or the "Save+Run" which obviously fails for non-code files.

This PR adds a simple "Save" button to the menu bar:

https://user-images.githubusercontent.com/114300/233834731-62f8b9d2-683f-4c2a-b577-6aadc709a2eb.mp4

The button is available in the mobile and the normal menu bar and only becomes visible once the editor has loaded a file.


makermelissa (approved):
Thanks. I think the Save + Run was originally intended to be the save option, but if you're working on a file not named code.py, this makes sense if you don't want to run it.

PR #162, opened 2024-02-20
Fix file listing for CP 9

makermelissa:
Fixes https://github.com/adafruit/circuitpython/issues/8875. This fixes the connection issue.

### DIFF
--- a/index.html
+++ b/index.html
@@ -312,7 +312,7 @@
             <tbody>
                 <tr>
                     <td>Board:</td>
-                    <td><a id="board"></a></td>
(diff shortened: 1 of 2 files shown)


tannewt (approved):
I think this is fine for now but it's probably better to use the version number because changes in the future may not be as simple.

makermelissa:
Fair enough. I wanted to just get it fixed so it works in both CP 8 and 9 for the moment. I would like to make better use of the additional info in a future PR and I can have it check the version number at that point.

PR #166, opened 2024-03-12
Fix JSZip import, so files can download

makermelissa:
This fixes the `Uncaught TypeError: JSZip is not a constructor` error that was happening when attempting to download a file. I'm not sure when this started happening, but changing the import fixed the issue.


makermelissa:
Merged in immediately because change is small and it unblocks me.

PR #200, opened 2024-06-17
Revamped USB file transfer using REPL

makermelissa:
Here is the PR that makes use of the updated REPL JS library to add REPL File Transfer.


dhalbert (approved):
I tested the latest on Windows and Linux, with an ESP32 and and ESP32-C3. It is working for me on Linux now (compared with a couple of days ago).

One slightly odd thing is that on Windows, if I shift-reload the editor, it seems to force hard reset of the board. (I was editing boot_out.txt as a test, and my changes were disappearing, which makes sense because that file is rewritten on a hard reset.)

PR #209, opened 2024-07-03
Display message about USB workflow issue

makermelissa:
As discussed in meeting. I opted for a popup because the functionality was already there.


dhalbert (approved):
This should be fine.

PR #211, opened 2024-07-08
Update REPL version to improve usb connection issues

makermelissa:


dhalbert (approved):
Thanks for debugging this.

PR #215, opened 2024-07-17
Use newest REPL JS Library

makermelissa:
Updated to reflect changes required for newer REPL library. This also fixes some minor issues such as appropriately updating the UI if the board is disconnected and updating any deprecated sass syntax.


dhalbert (approved):
:heavy_check_mark:

PR #216, opened 2024-07-17
Revert "Use newest REPL JS Library"

makermelissa:
Reverts circuitpython/web-editor#215. I did tons of testing locally and it was working perfectly. Now that it's live, it's not working at all. I think we need to set up a beta site as testing locally doesn't appear to be good enough.


dhalbert (approved):

makermelissa:
Huh, it seems to be working now...

PR #220, opened 2024-07-25
Use the latest REPL JS Library

makermelissa:
Improved reliability on freshly booted board by detecting if board does not have a prompt showing yet.


makermelissa:
Going to just merge myself since this is the beta site anyways and it's just a version bump.

PR #223, opened 2024-07-31
Implement multiple file selection dialog

### LINKED ISSUE #167: Add File Dialog Features related to selection
The file dialog is working great and pretty much as intended, but a couple of other features would really make it shine.

**Multiple File Selection**
It would be nice to be able to hold down Shift and/or Command/Ctrl to select multiple files (I like how it works better in Windows than Mac personally). Then the following operations could be done:
Delete, Download, Move

**Deselect**
This would probably go hand-in-hand with multiple file selection and would yield the same results as when a folder is first entered.

**Refresh**
This would simple retrieve the list of files again. It would probably be easiest to deselect the files, but perhaps a list could be compared and if the currently selected files still exist, maintain the selection.

makermelissa:
Fixes #167. This allows selecting multiple files in a similar way as OS's like windows allow. Then I had to update the behavior to handle multiple files such as downloading, moving, and deleting. Currently copying isn't implemented, but it would be for a different PR.

At this point, it appears to be working, but would like to do further testing before getting this merged in.

### DIFF
--- a/js/common/file_dialog.js
+++ b/js/common/file_dialog.js
@@ -436,9 +436,8 @@ class FileDialog extends GenericModal {
         }
 
         for (let filename of filenames) {
-            await this._showBusy(this._fileHelper.delete(filename));
             // Delete the item
-            await this._showBusy(this._fileHelper.delete(this._currentPath + filename));
+            await this._showBusy(this._fileHelper.delete(filename));
         }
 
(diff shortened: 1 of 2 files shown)


makermelissa:
I've been testing a bunch and it seems to consistently be working, so will merge this in for now since it's just the beta site.

Issue #32, opened 2022-10-19
Make current path in file dialog more obvious


makermelissa:
Make current path in file dialog more obvious what it is. When you first go there, it only shows a / and isn't very noticable until you navigate. Possibly prepend board name to it.

makermelissa:
I've been thinking about this and in cases where there is no CIRCUITPY drive, there's obviously no drive name. However, we could show something like [Root Folder] or perhaps the device type.

makermelissa:
Ok, after some experimenting, it appears adding a folder icon helps a lot.

makermelissa:
Fixed by #225 and #314.

PR #225, opened 2024-08-02
Show more info for all workflows

### LINKED ISSUE #224: Show additional userful info
We can show the type of device that the user is connected to and perhaps whether it is CircuitPython or MicroPython. We could also show the kind of workflow that is connected.

makermelissa:
Fixes #224.


makermelissa:
It was failing due to a bug in npm. I updated the workflow to get around the issue by removing `node_modules` and `package-lock.json`, then running `npm install`, then continuing as usual.

makermelissa:
Fixes #32 as well.

PR #242, opened 2024-08-28
save callback on this, and await folder functions

FoamyGuy:
This chang fixes an issue that is occurring for me in the beta version of the editor. Both the deployed beta page, and the beta branch when running locally behave the same for me and they're currently having this Javascript error when I click the 'Select Host Folder' or 'Use Host Folder' buttons 

```
index.js:94 Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'getFileClient')
    at HTMLButtonElement._selectHostFolder (index.js:94:3384)
```
I added some additional console logging and found that `this.fileHelper` was indeed undefined when it tried to be accessed by this callback. 

I looked into the changes from main to beta and saw that these callbacks used to be wrapped in the anonymous async functions and awaited so I tried putting that back and it seems to resolve the issue. 

I also store the callback so that it can later be passed to `removeEventListener()` This is very similar to what ended up being a fix for the duplicated serial output, I'll submit a separate PR for that fix though as its unrelated to this JS error I believe.

I tested this version locally and found that with it I am now able to select the host folder successfully and carry on using the editor.

### DIFF
--- a/js/workflows/usb.js
+++ b/js/workflows/usb.js
@@ -21,6 +21,9 @@ class USBWorkflow extends Workflow {
         this._partialToken = null;
         this._uid = null;
         this._readLoopPromise = null;
+        this._btnSelectHostFolderCallback = null;
+        this._btnUseHostFolderCallback = null;
+
     }
 
     async init(params) {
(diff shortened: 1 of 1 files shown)


makermelissa (approved):
Thank you

PR #243, opened 2024-08-28
Fix duplicate serial console messages

### LINKED ISSUE #234: Duplicate Serial Input & Output After Disconnect/Reconnect
### CircuitPython Version

```python
Adafruit CircuitPython 9.1.1 on 2024-07-22; Adafruit Feather RP2040 with rp2040
Board ID:adafruit_feather_rp2040
```
(but seems to affect all versions equally)
```


### Workflow(s)

USB

### Browser(s)

```python
Version 127.0.6533.99 (Official Build) (64-bit)
```

### Console Log

```python
Read Loop Init
index.js:96 Read Loop Started
index.js:92 Normal Prompt not detected.
index.js:94 Getting Device UID...
index.js:96 Device UID: K4545373038053531
454537303805353
index.js:94 Successfully loaded directory: TOUCH_DECK
index.js:94 New folder name: TOUCH_DECK
index.js:94 SerialPort {onconnect: null, ondisconnect: null, readable: ReadableStream, writable: WritableStream}
index.js:92 Connected!
```

### Steps

1. Connect device using USB workflow
2. Open serial pane and press ctrl-c / ctrl-d. Observe single copy of any outputs from code.py + the repl prompt
3. (optional) type into the repl, observe single instance of typed characters appearing
4. Press Disconnect button near top right
5. Press Connect button near top right and complete the USB workflow connection / pairing process again
6. Open serial pane and press ctrl-c / ctrl-d. Observe duplicated copy of outputs from code.py and the repl prompt
7. (optional) type into the REPL, observe duplicated instances of typed characters appearing

You can continue to repeat the steps to add an additional duplicated instance of the input and ouput each time you disconnect / reconnect. 

Reloading the page with F5 and then reconnecting will get you back to a single non-duplicated instance of input / output

### Description

_No response_

### Additional information

My guess without looking further into the code is a JS callback registered for serial connection that doesn't get unregistered when disconnect happens, so it sticks around and a new callback gets registered upon re-connecting leading to the duplicates.

FoamyGuy:
Resolves #234 

I am not positive, but I think the existing `removeEventListener()` is not working because `this.onSerialReceive.bind(this)` is producing a new function each time it's called. So when it gets called for remove() technically it is a different instance of the function than the one that ends up being passed to `addEventListener()` and thus the one that will need to be passed next time to remove in order for it to get de-activated.

This resolves it by creating the callback once, storing it on `this` and then referencing it from there for both the add and remove listener calls.

I tested this locally and it does seem to resolve the duplicate serial issue, with it I can now disconnect / re-connect multiple times and still end up with just a single instance of the serial output.

### DIFF
--- a/js/workflows/usb.js
+++ b/js/workflows/usb.js
@@ -237,8 +237,8 @@ class USBWorkflow extends Workflow {
     async _switchToDevice(device) {
 
         device.removeEventListener("message", this.messageCallback);
-        this.messageCallback = this.onSerialReceive.bind(this);
-        device.addEventListener("message", this.messageCallback);
+        this._messageCallback = this.onSerialReceive.bind(this);
+        device.addEventListener("message", this._messageCallback);
 
         let onDisconnect = async (e) => {
             await this.onDisconnected(e, false);
--- a/js/workflows/usb.js
+++ b/js/workflows/usb.js
@@ -237,7 +237,7 @@ class USBWorkflow extends Workflow {
     // Workflow specific Functions
     async _switchToDevice(device) {
 
-        device.removeEventListener("message", this.messageCallback);
+        device.removeEventListener("message", this._messageCallback);
         this._messageCallback = this.onSerialReceive.bind(this);
         device.addEventListener("message", this._messageCallback);
 
--- a/js/workflows/usb.js
+++ b/js/workflows/usb.js
@@ -21,6 +21,7 @@ class USBWorkflow extends Workflow {
         this._partialToken = null;
         this._uid = null;
         this._readLoopPromise = null;
+        this._messageCallback = null;
     }
 
     async init(params) {
--- a/js/workflows/usb.js
+++ b/js/workflows/usb.js
@@ -235,8 +235,10 @@ class USBWorkflow extends Workflow {
 
     // Workflow specific Functions
     async _switchToDevice(device) {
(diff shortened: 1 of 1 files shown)


js/workflows/usb.js
```diff
@@ -235,8 +235,10 @@ class USBWorkflow extends Workflow {
 
     // Workflow specific Functions
     async _switchToDevice(device) {
-        device.removeEventListener("message", this.onSerialReceive.bind(this));
-        device.addEventListener("message", this.onSerialReceive.bind(this));
+
+        device.removeEventListener("message", this.messageCallback);
+        this.messageCallback = this.onSerialReceive.bind(this);
+        device.addEventListener("message", this.messageCallback);
```
makermelissa, line 241:
```suggestion
        this._messageListener = this.onSerialReceive.bind(this);
        device.addEventListener("message", this._messageListener);
```

makermelissa (changes requested):
Let's make it private and since it's not actually a callback, let's rename it to this._messageListener. Also, please add to the constructor.

makermelissa:
Updated my comment as using "Callback" was consistent with other usages.

FoamyGuy:
Thank you, It's been renamed, and added to the constructor with latest commits.

makermelissa (approved):
Thanks

Issue #244, opened 2024-08-29
Serial Plotter?


FoamyGuy:
Is there interest in having a serial plotter similar to the one inside of Mu?

makermelissa:
Possibly. I believe I did one in another project (https://github.com/adafruit/Adafruit_WebSerial_Plotter) that could possibly be ported here.

FoamyGuy:
Nice, thank you for the link, I hadn't seen that before. 

I started tinkering with this idea a bit this morning, I've gotten  as far as making a spot in the layout for it, and a button that reveals/hides it but no actual plotting functionality yet. I'm planning initially on putting it in the lighter grey box within the serial-page portion of the layout, but am open to feedback or other suggestions if you have different ideas about how or where to integrate it into the UI.

![image](https://github.com/user-attachments/assets/0c051340-04a6-4b91-a323-944983a1edb1)

I've used Apex Charts (https://apexcharts.com/) for a few things recently and was thinking of using it for this plotter. But I have also used Chart.js in the past which it looks like the older serial plotter you worked on uses. I may opt for Chart.js instead and see how much of that older version I can re-use. 

I'm assuming that the XT and XY plot types are like "value over time" vs. "value over other value". I hadn't considered that option before but it does seem like nice functionality. 

The buffer size is nice as well, I thought about it some but hadn't settled on any solution beyond having the clear button also clear out the chart data in addition to the terminal output.

makermelissa:
Fixed by #250 and #314.

PR #245, opened 2024-08-30
Serial plotter implementation

FoamyGuy:
resolves: #244 

This is the start of implementing the serial plotter functionality.

It's coming along nicely, here is a screen recording and picture:
[Screencast from 2024-08-30 13-02-27.webm](https://github.com/user-attachments/assets/d0c93a01-2db2-4d1e-95ed-7dd76fcca7d5)
Screenshot from 2024-08-30 13-06-48

I'm leaving this marked as draft for now though because I do think it still needs some work before being ready. I'm open to any feedback on code or behavior / visual layout. 

One of the remaining problems to be solved is that the portion of the terminal that is shown is at the very top of the terminal pane, so as new things are printed they don't actually become visible while the plotter is showing until enough items have been printed to make it scroll enough to actually start showing values. Once they do become visible they are also outdated as well with several newer prints not currently visible below them working to push them up.

I am thinking that it may require changing the number of lines available in the terminal scroll pane while the serial plotter is showing, but haven't explored that further than having the idea yet.

There is probably opportunity to refactor more of the plotter functionality into the new plotter.js file as well. 

Right now only the line chart over time is implemented, but there is a commented out dropdown that is intended to eventually allow the user to select scatterplot.


FoamyGuy:
The latest commit resolves the issue of the terminal showing only the top portion of data. We already are using the addon-fit module which handles it perfectly it just needs to be called when the plotter is hidden and shown. 

So now the most recently printed values are visible within the terminal pane while the plotter is showing.

FoamyGuy:
I think the functionality and code for this are ready for review. 

However I do see that the actions are failing now, and I'm not entirely sure why. The error reported pertains to `npm ci` 

I found that I got roughly the same error if I run `npm ci` locally. I tried `npm install` locally like the error message suggested and that does seem to resolve the issue, I can run `npm ci` afterward and it will succeed.  But it seems to have made further changes inside of package-lock.json that are unrelated to the plotter and seem to be undoing some of the changes that were brought in when I merged `beta` into this branch. 

At one point I also accidentally merged `main` into this branch instead of `beta` as intended. So that could have been part of what has caused issues with the ci / actions for this branch. 

I'm sure that it is possible to repair this branch somehow but I feel so far that I've been failing to do so and making it more mucked up.  I think it might be easiest for me to just make a new branch started fresh from current beta and open a new PR with it.

Issue #285, opened 2025-01-22
Extra empty lines in files read via USB serial


FoamyGuy:
### CircuitPython Version

```python
Adafruit CircuitPython 9.2.3 on 2025-01-17; Adafruit Mini Sparkle Motion with ESP32
```

### Workflow(s)

USB

### Browser(s)

```python
Google Chrome Version 132.0.6834.83 (Official Build) (64-bit)
```

### Console Log

```python

```

### Steps

1. Connect to ESP32 based device using USB mode (it will have to use repl file transfer)
2. open a file from the device
3. File shown in the editor has extra empty lines inserted every other line.

### Description

When reading via USB the files come accross with extra empty lines inserted i.e. 

```
import board
import neopixel

from adafruit_led_animation.animation.comet import Comet
from adafruit_led_animation.animation.rainbow import Rainbow
from adafruit_led_animation.color import GREEN
```

becomes:

```
import board

import neopixel



from adafruit_led_animation.animation.comet import Comet

from adafruit_led_animation.animation.rainbow import Rainbow

from adafruit_led_animation.color import GREEN
```

### Additional information

Using the web mode instead of USB on the same device and browser works correctly and the file is shown normally without the extra lines inserted.

FoamyGuy:
resolved by #286

PR #286, opened 2025-01-22
remove \r from file contents when read via USB repl

FoamyGuy:
@ladyada 
Resolves: #285

When read via usb repl the file contains both `\r\n` and the code editor renders both as empty lines. This change strips the `\r` leaving only a `\n` to be rendered. 

Tested successfully on a locally running instance. 
image


makermelissa (approved):
Thanks. I can't remember why I added this before, but if it fixes the issue, then good.

Issue #289, opened 2025-01-28
serial status line missing on beta


FoamyGuy:
### CircuitPython Version

```python
Adafruit CircuitPython 9.2.3 on 2025-01-17; Adafruit Sparkle Motion with ESP32
```

### Workflow(s)

USB

### Browser(s)

```python
Google Chrome 132.0.6834.110
```

### Console Log

```python

```

### Steps

connect board with USB workflow

open serial pane

### Description

Live version:
![Image](https://github.com/user-attachments/assets/b9a19cb8-5377-4c3e-9b3e-2058339e064f)

Beta version:
![Image](https://github.com/user-attachments/assets/a6db2d2f-176f-4d43-8e93-43ed2400e47f)

### Additional information

The status information does blink visible momentarily I think. Probably something about the plotter button has broken it.

FoamyGuy:
Maybe Wifi Off vs. On plays into this somehow. Currently I am seeing the status info on the beta page. My wifi is currently off. 

I'm not positive what state it was in when the issue was created and the screenshot showing it missing was taken, but it's possible that it was on for that screenshot.

makermelissa:
I just tried beta and it shows for me. I suspect the issue was actually the new plotter button made the contents of the serial pane a bit wider, which makes the status disappear due to not having enough room.

makermelissa:
Closing as this is the intended behavior.

PR #291, opened 2025-01-31
Fix file download issues

### LINKED ISSUE #290: USB REPL File Mode - Download Files Not Working
### CircuitPython Version

```python
Adafruit CircuitPython 9.2.3 on 2025-01-17; Adafruit Sparkle Motion with ESP32
```

### Workflow(s)

USB

### Browser(s)

```python
Chrome Version 132.0.6834.110 (Official Build) (64-bit)
```

### Console Log

```python
index.js:20 Uncaught (in promise) TypeError: Failed to execute 'createObjectURL' on 'URL': Overload resolution failed.
    at Function.l [as saveAs] (index.js:20:20083)
    at Pd._download (index.js:31:101955)
    at async Pd._handleDownloadButton (index.js:31:101145)
```

### Steps

1. connect to board with USB workflow
2. press open button
3. select a single file
4. press download button

### Description

Download does not complete and error is printed in console. 

### Additional information

If you select multiple files it does not have the same error in the console but the resulting downloaded zip contains 0 byte files.

makermelissa:
Fixes #290. This also updates the version of circuitpython-repl-js with some fixes in it that were added in https://github.com/adafruit/circuitpython-repl-js/pull/15.

### DIFF
--- a/js/common/file_dialog.js
+++ b/js/common/file_dialog.js
@@ -548,20 +548,14 @@ class FileDialog extends GenericModal {
         input.click();
     }
 
-    // Currently only files are downloadable, but it would be nice to eventually download zipped folders
     async _handleDownloadButton() {
-        // TODO: Implement a way to download multiple files at once into a zip file
-
-        await this._download(this._getSelectedFilesInfo());
+        await this._showBusy(this._download(this._getSelectedFilesInfo()));
     }
 
     async _download(files) {
         if (!this._canDownload()) return;
 
-        let folder, blob, filename;
-
-        // If we only have 1 item and it is a file, we can download it directly
-        // Otherwise, we need to zip the files and download the zip keeping the structure intact
+        let blob, filename;
 
         // Function to read the file contents as a blob
         let getBlob = async (path) => {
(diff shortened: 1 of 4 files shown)


makermelissa:
I think this may also fix #289 as status works when running locally on my machine.

FoamyGuy:
I am trying to run this branch locally but I am still having trouble downloading files and folders. 

If I select a single file like code.py and click the download button I see the loading indicator spin, then it goes away but no file is downloaded. In the console it prints this error message:
```
Uncaught (in promise) TypeError: Failed to execute 'createObjectURL' on 'URL': Overload resolution failed.
    at g (file-saver.js?v=b3c4113b:47:224)
    at FileDialog._download (file_dialog.js:614:9)
    at async FileDialog._handleDownloadButton (file_dialog.js:555:9)
```

If I instead select a folder like `lib/` and click the download button it does download a file `lib.zip` but the file is only 22 bytes and my OS reports that it's an empty archive when I attempt to extract it. I do have 3 .mpy files inside of lib on the device but they don't end up inside of the zip for me. 

I'm not sure if I understand fully how version numbers work with NPM modules. I saw in this change that adafruit/circuitpython-repl-js  is moved to version `3.2.2`. After pulling this branch and running `npm install` I noticed that inside of `node_modules/@adafruit/circuitpython-repl-js/package.json` I see `"version": "2.0.1",`. Does this mean that I am not successfully getting the version specified in this branch?

FoamyGuy:
I have found and run `npm update` and then `npm install` again. After that I am now able to successfully download single files like `code.py` however downloading a folder like `lib/` is still resulting in an empty zip. there are no errors or other console output during the folder download attempt.

makermelissa:
Ok, I don't think I tested that, so I'll take a look.

makermelissa:
It seems like every time I fix one thing another thing breaks. I'm getting close to finishing and realized there are 5 different test cases that need to all be working:

- Single file selected
- Root Folder (no files selected)
- Sub Folder (no files selected)
- Subfolder Selected
- Multiple Files Selected

makermelissa:
Ok, lots of testing later, it should all be working now. I ended up rewriting a lot of it and it makes a lot more sense now.

FoamyGuy (approved):
I tested this locally with a Sparkle motion mini. Downloading a single file, a folder, and the entire drive are all working correctly for me now. 

Selecting multiple files and then pressing download did download a zip but the files inside were 0 bytes. However I didn't realize that was intended to be supported, so I see it as a bonus if it does ultimately work. 

Another thing that might be nice to have ultimately is more specific status during the downloading, if it can show each filename as it's working on them it will help assure the user that "it's working" and not stalled.

makermelissa:
@FoamyGuy did you run `npm install` again? I updated the circuitpython-repl-js package again.

makermelissa:
I'm going to merge this anyway because my testing produced correct results and I believe the beta site will also have correct results.

PR #292, opened 2025-02-07
Show busy while user selects port and handle canceling properly

### LINKED ISSUE #240: Have a new dialog message appear on screen while board is connecting
Currently when a user tries to connect a board via Web Serial, the same dialog box (Select Serial Device - Connect serial device) remains on screen after a user clicks connect and selects their COM port. This is confusing to users because they aren't sure if a connection is in progress or has failed. The user can also keep clicking the same series of prompts with no feedback. I think it would be helpful to let a user know that the editor is trying to connect to their board that they selected or to tell them that a connection failed.

makermelissa:
Fixes #240.

### DIFF
--- a/js/workflows/usb.js
+++ b/js/workflows/usb.js
@@ -125,7 +125,12 @@ class USBWorkflow extends Workflow {
             // This would help with other workflows as well
         } else {
             console.log('Requesting any serial device...');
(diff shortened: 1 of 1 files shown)


FoamyGuy (approved):
This looks good to me. I tested locally with a mini sparkle motion. Thank you!

PR #296, opened 2025-02-12
Resize fix

makermelissa:
This is to address #21.

After fiddling with the terminal fitter addon for a while, I discovered it made more sense to write my own custom function that wasn't buggy as I already had to write a large portion of it to get around the bugs and it didn't make sense to update the fitter addon code because the solution was very tailored to the code editor.

### DIFF
--- a/js/layout.js
+++ b/js/layout.js
@@ -15,6 +15,9 @@ const SETTING_TERMINAL_VISIBLE = "terminal-visible";
 const UPDATE_TYPE_EDITOR = 1;
 const UPDATE_TYPE_SERIAL = 2;
 
+const MINIMUM_COLS = 2;
+const MINIMUM_ROWS = 1;
+
 function isEditorVisible() {
     return editorPage.classList.contains('active');
 }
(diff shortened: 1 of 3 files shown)


js/layout.js
```diff
@@ -108,13 +113,40 @@ function updatePageLayout(updateType) {
...
+                const TERMINAL_COL_WIDTH = state.terminal._core._renderService.dimensions.css.cell.width;
+
+                // Get the height of the header, footer, and serial bar to determine the height of the terminal
+                let siteHeader = document.getElementById('site-header');
+                let mobileHeader = document.getElementById('mobile-header');
+                let headerHeight = siteHeader.offsetHeight;
+                if (siteHeader.style.display === 'none') {
+                    headerHeight = mobileHeader.offsetHeight;
+                }
+                let foorterBarHeight = document.getElementById('footer-bar').offsetHeight;
```
tannewt, line 133:
```suggestion
                let footerBarHeight = document.getElementById('footer-bar').offsetHeight;
```

tannewt (approved):
One variable name typo you may want to fix. Looks good otherwise. Thanks!

js/layout.js
makermelissa: Good catch. I also need to update line 136.

Issue #43, opened 2022-12-04
Handle partial connections better


makermelissa:
For USB and possibly BLE, we require the user to perform at least 2 steps to connect. If the user cancels in the middle of this, I'm not sure if this is being handled properly. Additionally, the UI could use some improvement to better inform the user that a step is complete. 
I don't think this affects Web Workflow because it's basically 1 step, though I suppose if the user fails to type in credentials it may not be handled properly, so that should probably be checked too, though I imagine the result is they'll either get an error or it will prompt them again next time. 

Here's what I think needs to be fixed for this issue to be complete:

- [x] If the user closes the dialog early, a disconnect should be performed
- [x] Check that Web Workflow is properly handled if the credentials dialog is dismissed
- [ ] As the user completes steps, something like a checkmark should visually inform the user that it is complete
- [ ] If the instructions are on a small screen like a phone, it should scroll to the next step the user is expected to do
- [ ] Fix it if there is an issue so that the user knows what to do

makermelissa:
Fixed via https://github.com/circuitpython/web-editor/pull/298

PR #298, opened 2025-02-17
Improve connection step updating and add checkmarks

makermelissa:
This completes another step in #43. While adding code to mark steps complete, I was able to abstract the function to enable/disable buttons.

### DIFF
--- a/images/checkmark.svg
+++ b/images/checkmark.svg
@@ -1,0 +1,1 @@
+<svg viewBox="0 0 50 50" xmlns="http://www.w3.org/2000/svg"><path d="m46 14-21 21.6-7-7.2-7 7.2 14 14.4 28-28.8z" fill="#57c84d"/></svg>
--- a/js/workflows/ble.js
+++ b/js/workflows/ble.js
@@ -29,6 +29,12 @@ class BLEWorkflow extends Workflow {
         this.infoDialog = new DeviceInfoModal("device-info");
         this.partialWrites = true;
         this.type = CONNTYPE.Ble;
+        this.buttonStates = [
+            {reconnect: false, request: false, bond: false},
+            {reconnect: false, request: true, bond: false},
+            {reconnect: true, request: true, bond: false},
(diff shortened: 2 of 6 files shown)


tannewt (commented):
Would this work with an emoji check instead of an svg?✔️ 

Also, could you add a short screen capture to show what this looks like? Thanks!

makermelissa:
Sure, I could give it a try. It should work in theory. I'll see how it looks though.

makermelissa:
I think the SVG looks better. Here's a comparison:
Screenshot 2025-02-18 at 3 45 56 PM
Screenshot 2025-02-18 at 3 42 01 PM

Here' a screen recording to see it in action:
https://github.com/user-attachments/assets/02e40b89-ebab-423c-8d1b-8614da894eaa

tannewt:
Thank you for the screen recording! Do we need to show step 1 at that point at all? Why not just show step 2? Earlier it only shows step 1.

makermelissa:
It allows users to still connect to a different device in case they selected the wrong one or to go forward with step 2.

tannewt:
I'm not convinced you need to do that here. They can always disconnect and click connect again.

If you do want to show it still, then I'd update the step 1 info with what was selected. That way one can tell if the correct thing was selected.

makermelissa:
After trying out some things, I think you're correct. I updated the code to disable the Connect to Device button. I don't want to remove the step altogether because the CSS  would automatically renumber the steps and step 2 would appear to be step 1.

dhalbert:
One thing I noticed recently is that when the UF2 is downloaded and then uploaded to the board, the progress bar just says "Downloading": it's not two steps. Or am I missing something?

dhalbert:
> ~~One thing I noticed recently is that when the UF2 is downloaded and then uploaded to the board, the progress bar just says "Downloading": it's not two steps. Or am I missing something?~~

Sorry, never mind, thought this was the web installer.

tannewt (approved):
Ok, thanks!

Issue #183, opened 2024-04-30
Unable to access the Open diaglog when connecting via bluetooth (hangs on open and save with spinner)


wuftymerguftyguff:
### CircuitPython Version

```python
Adafruit CircuitPython 8.2.10 on 2024-02-14; Adafruit ItsyBitsy nRF52840 Express with nRF52840
Board ID:itsybitsy_nrf52840_express
UID:1027F3CECE5BC1D4
```

### Workflow(s)

BLE

### Browser(s)

```python
Version 124.0.6367.78 (Official Build) (x86_64) (mac + windows)
```

### Console Log

```python
﻿
index.js:37 Load different workflow
context-menu-content-scripts.js:25 Using workaround to get event.path
index.js:35 Getting existing permitted Bluetooth devices...
index.js:35 > Got 1 Bluetooth devices.
index.js:35 Watching advertisements from "CIRC"...
index.js:35 > Received advertisement from "CIRC"...
index.js:35 Connecting to GATT Server from "CIRC"...
index.js:35 > Bluetooth device "CIRC connected.
index.js:35 
BluetoothDevice {id: 'WVH+/a5nYxWRtA9KJo1lqQ==', name: 'CIRC', gatt: BluetoothRemoteGATTServer, ongattserverdisconnected: null, watchingAdvertisements: false, …}
index.js:35 connected 
BluetoothRemoteGATTServer {device: BluetoothDevice, connected: true}
index.js:35 
(2) [BluetoothRemoteGATTService, BluetoothRemoteGATTService]
index.js:35 Initializing File Transfer Client...
index.js:35 DOMException: GATT operation failed for unknown reason. undefined
index.js:35 Connected!
2
context-menu-content-scripts.js:25 Using workaround to get event.path
index.js:21 caught connection error DOMException: GATT operation failed for unknown reason. undefined
index.js:21 caught write error TypeError: Cannot read properties of null (reading 'writeValueWithoutResponse')
    at wS._write (index.js:21:6026)
    at wS.listDir (index.js:21:12322)
    at async ZS.showBusy (index.js:35:106432)
    at async Za._openFolder (index.js:35:95507)
    at async Za.open (index.js:35:95119)
    at async ZS.openFileDialog (index.js:35:108201)
    at async ZS.openFile (index.js:35:107915) TypeError: Cannot read properties of null (reading 'writeValueWithoutResponse')
    at wS._write (https://code.circuitpython.org/assets/js/index.js:21:6026)
[...]

PR #300, opened 2025-03-02
Improvements to Bluetooth connectivity

makermelissa:
This is to address #183. This improve Bluetooth workflow in several ways:

- Error messages are now displayed as a popup rather than in the console, so the user is aware that there is a problem.
- When a problem with connecting happens, a solution of removing the device from the OS is now displayed, which works consistently
- The workflow is a bit rearranged so that there are fewer problems connecting
- A new more specific flag (about://flags/#enable-web-bluetooth-new-permissions-backend) is recommended over the more general (about://flags/#enable-experimental-web-platform-features) flag.
- https enabled in vite because web bluetooth is supposed to require that (it wasn't before due to the more general flag, which I believe was part of the issue we were having with Bluetooth.


tannewt (approved):
Thank you!

PR #317, opened 2025-06-25
Fix MicroPython showing busy forever

makermelissa:
This fixes an issue with the busy indicator staying up indefinitely when connecting to MicroPython. This like affects devices like the ESP32 as well. This was fixed by moving the busy indicator to only the functions that take more time like the fileops functions.


FoamyGuy (approved):
Thank you!

makermelissa:
@Dan, it never finishes until the board is disconnected. It enters a read loop to process serial data.

PR #329, opened 2025-07-12
[DNM for discussion] Conditionally change button text on USB connect. Resolves #328

### LINKED ISSUE #328: Minor UI issue: On windows, connecting to filesystem shows filesystem path as "/"
### CircuitPython Version

```python
Adafruit Circuit Python 9.2.8
```

### Workflow(s)

USB

### Browser(s)

```python
Version 138.0.7204.98 (Official Build) (64-bit)
```

### Console Log

```python
Initializing File Transfer Client...  index.js:116 
New folder name: \
```

### Steps

1. Connect a board that has native USB filesystem support (like a RP2040 board, or an Adafruit Metro M7)
2. Connect to the board.
3. After selecting the port in step 1, it asks you to pick the path to the CircuitPython Device
4. It puts up a button that says "Use \" if everything goes well on Windows.

### Description

There is a button that says, "Use \\" that you have to click on:

Image

This is very confusing to Windows Users.  I realize on Unix filesystems it would show something unique, when you select a drive, but on Windows there may be many "\\" directories, one for each drive mounted.  

Reading through the documentation, the drive letter and other things to help the browser are intentionally omitted for security. IMHO, it would be better to just show an "OK" button (or to just accept the selection without additional confirmation?)

### Additional information

This is quite minor, but I thought it might be a good starter bug for someone (maybe me)

ericzundel:
Changes the text in the UI button for selecting a filesystem mount point based on platform.  Putting the pathname on the filesystem could be confusing for Windows or ChromeOS users where they don't have the concept of a single filesystem tree the way *NIX filesystems are used to.
 
 
image

Resolves #328

### DIFF
--- a/index.html
+++ b/index.html
@@ -308,7 +308,7 @@
                     <h1>Select USB Host Folder</h1>
                     <p>Select the root folder of your device. This is typically the CIRCUITPY Drive on your computer unless you renamed it. If your device does not appear as a drive on your computer, it will need to have the USB Host functionality enabled.</p>
                     <p>
-                        <button class="purple-button hidden" id="useHostFolder">Use <span id="workingFolder"></span></button>
+                        <button class="purple-button hidden" id="useHostFolder"><span id="workingFolder"></span></button>
                         <button class="purple-button first-item" id="selectHostFolder">Select New Folder</button>
                     </p>
                 </div>
--- a/js/common/utilities.js
+++ b/js/common/utilities.js
@@ -56,6 +56,25 @@ function isLocal() {
     return (isMdns() || location.hostname == "localhost" || isIp()) && (location.pathname == "/code/");
 }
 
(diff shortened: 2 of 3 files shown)


ericzundel:
I'm puting up this change mostly just to start a discussion.  Do we want the pathname in this button?   For example, on ChromeOS and Windows we could just make it say "OK" but then, users who don't dig down in the filesystem might be confused by the path that comes up.

FYI, I have only tested sucessfully on my windows laptop at home.  I'm wondering what it shows on a  Mac? Underneath it all is a single UNIX filesystem, but users that don't use the shell don't see that.

makermelissa (approved):
Thank you. This will work for the time being. We can always improve on it later if we figure out a way to show the windows/chrome folder.

Issue #327, opened 2025-07-11
Hang when connecting to a RP2040 with filesystem mounted read-only


ericzundel:
### CircuitPython Version

```python
CircuitPython 9.2.8
Raspberry Pi Pico
```

### Workflow(s)

USB

### Browser(s)

```python
Google Chrome Version 138.0.7204.98 (Official Build) (64-bit)
Windows 11
```

### Console Log

```python
index.js:32 Uncaught (in promise) TypeError: Cannot read properties of null (reading 'readOnly')
    at Wu.open (index.js:32:93023)
    at RR.saveFileDialog (index.js:112:6199)
    at RR.saveFileAs (index.js:112:5645)
    at HTMLButtonElement.<anonymous> (index.js:118:11847)


When trying to recreate this, I got some different errors.  The connect attempt suceeded, but saving hung with these errors in the console:

index.js:118 write failed InvalidStateError: An operation that depends on state cached in an interface object was made but the state had changed since it was read from disk. undefined
sy	@	index.js:118
await in sy		
saveFile	@	index.js:112
jb	@	index.js:118
await in jb		
(anonymous)	@	index.js:118
index.js:118 write failed TypeError: Cannot read properties of undefined (reading 'split')
    at PR._splitPath (index.js:115:3146)
    at PR.writeFile (index.js:115:2960)
    at async cT.writeFile (index.js:93:8639)
    at async RR.showBusy (index.js:112:4411)
    at async RR.writeFile (index.js:112:6345)
    at async sy (index.js:118:16147) TypeError: Cannot read properties of undefined (reading 'split')
    at PR._splitPath (https://code.circuitpython.org/assets/js/index.js:115:3146)
    at PR.writeFile (https://code.circuitpython.org/assets/js/index.js:115:2960)
    at async cT.writeFile (https://code.circuitpython.org/assets/js/index.js:93:8639)
    at async RR.showBusy (https://code.circuitpython.org/assets/js/index.js:112:4411)
    at async RR.writeFile (https://code.circuitpython.org/assets/js/index.js:112:6345)
    at async sy (https://code.circuitpython.org/assets/js/index.js:118:16147)
sy	@	index.js:118
setTimeout		
sy	@	index.js:118
await in sy		
saveFile	@	index.js:112
jb	@	index.js:118
await in jb		
(anonymous)	@	index.js:118
```

### Steps

1. Connect the RP2040  device to the computer with USB. I'm using a Raspberry Pi Pico 2.
[...]

PR #334, opened 2025-07-26
Added status message area to USB connect workflow

ericzundel:
This PR attempts to provide user friendly feedback during the USB connection using WebSerial and selecting a filesystem mount point.

image

I worked on this PR when noticing some hiccups when trying to connect my microprocessor via USB.  See #327, although it's not a great solution to the read-only issue as the element I created to show status isn't visible after the connect part of the workflow finishes.

### DIFF
--- a/index.html
+++ b/index.html
@@ -301,6 +301,11 @@
                     </p>
                 </div>
             </section>
+            <section>
+                <div class="connect-status" hidden>
+                    <div class="connect-status-content"></div>
+                </div>
+            </section>
         </div>
     </div>
     <div class="popup-modal shadow closable" data-popup-modal="device-discovery">
--- a/index.html
+++ b/index.html
@@ -221,14 +221,21 @@
                 <div class="step-number"></div>
                 <div class="step-content">
                     <h1>Request Bluetooth Device</h1>
-                    <p>CircuitPython boards with <a href="https://circuitpython.org/downloads?features=Bluetooth%2FBTLE">nrf chips need
-                        CircuitPython 7.0.0 or newer</a>. The first time a device is connected to your host,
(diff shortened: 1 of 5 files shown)


ericzundel:
I just tried the Web workflow. This is mysterious to me!  I did get it to work after a few tries and filed issue 370.  However, I confess that I don't yet understand how the redirection to the microprocessor's port 80/code/ URL works.  Does the microprocessor proxy to http://code.circuitpython.org/ to get the UI?  I can see the UI as a part of this codebase, surely it isn't all compiled into circuitpython executable?  

For the purposes of this PR, It isn't clear to me that anything can go wrong in the current web workflow that can be made visible to the user. If you can't connect to the uP IP address:80/code/ you just get a timeout from the browser.

I'll look at BLE next.

makermelissa:
> I just tried the Web workflow. This is mysterious to me! I did get it to work after a few tries and filed issue 370. However, I confess that I don't yet understand how the redirection to the microprocessor's port 80/code/ URL works. Does the microprocessor proxy to http://code.circuitpython.org/ to get the UI?

Yes, that's exactly what it does. The device code contains [a very basic page](https://github.com/adafruit/circuitpython/blob/1a6c2b96db62dbe7e1dbec2b57d67f5778e20a69/supervisor/shared/web_workflow/static/code.html) and uses https://github.com/circuitpython/web-editor/blob/main/public/assets/js/device.js to bootstrap the rest of website. This makes it very maintainable and tricks the browser into thinking the code is running on the device because the devices don't currently support https for web workflow.

> I think I can add the status calls to the BLE workflow. I filed a couple of bugs, I'll try to submit any fixes for those I find along the way (like adding a link to the online documentation)

Thank you

PR #350, opened 2025-09-04
Streamlined BLE Connection Workflow

makermelissa:
This has been streamlined to make the flow much simpler. Here's a functional comparison of the flows:

**Before**
BLE Connect Flow (Before)

**After**
BLE Connect Flow (After)


tannewt (approved):
Thank you for the simplification!

PR #412, opened 2025-10-27
wip(wifi workflow): DELETE method available doesn't represent writable on CPY 10.0.3 lilygot tdisplay S3, switch to fs json like circup

tyeth:
This switches to using the file web api response to get the writable state, which is returned whenever the listDir api is used.

On the lilygo T-Display S3 the HTTP PUT failed due to not being writable, and the http response was something like "USB ACTIVE, try reseting board" or something like that, but the response text is not displayed in console nor to user.

Looking deeper, the PUT request should have been skipped, due to writable: false being part of the listDir response for "/".
Checking the current readOnly code, it checks for an HTTP OPTIONS query and inspects the returned methods for DELETE, which **was present** when my device was **not writable**. The JSON property was correctly returned.

This switches to using the /fs/ API check instead of relying on HTTP DELETE method being supported.

It's probably worth refactoring this into a class level flag thats updated with each listDir and then cached, until a disconnect event or failed write call at which point it's forcefully refreshed.

Calling like it currently does at each call of readOnly is actually fine (at least on a responsive wifi network).

### DIFF
--- a/js/common/web-file-transfer.js
+++ b/js/common/web-file-transfer.js
@@ -7,7 +7,11 @@ class FileTransferClient {
 
     async readOnly() {
         await this._checkConnection();
-        return !this._allowedMethods.includes('DELETE');
+        console.log("Checking read only");
+        const response = await this._fetch("/fs/", {method: "GET", headers: {"Accept": "application/json"}})
+        const result = await response.json();
+        //TODO: Tyeth: cache this value until reconnection, as listdir / connect already fetch it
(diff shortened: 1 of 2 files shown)


makermelissa:
Is this still in progress?

tyeth:
Hey @makermelissa, sorry this is no longer in progress, feel free to close it / take it on, as desired. Or I could schedule some time to revisit it if that would be preferable.

makermelissa:
I can take it on. Thanks.

Issue #355, opened 2025-09-05
User Settings Panel


makermelissa:
It would be great to have a user settings panel to be able to select theme, font size, etc.

makermelissa:
Fixed by #473 and #442.

Issue #354, opened 2025-09-05
Color Theme Selection


makermelissa:
This is related to https://github.com/circuitpython/web-editor/issues/181 but goes further:

The idea is the user could go to a settings page and make adjustments to a preselected set of themes. We could start with a light and dark there, but this would involve adding the scaffolding to have as many themes as we want.

makermelissa:
Fixed by #473 and #442.

Issue #181, opened 2024-04-22
Light mode


dhalbert:
I would appreciate a "light mode" display mode in addition to the current dark mode.

makermelissa:
Great idea. I tend to use editors in light mode too.

makermelissa:
I'm thinking some sort of predefined themes. I'll probably just start with current plus light mode.

makermelissa:
Fixed by #473 and #442.

PR #442, opened 2026-01-09
Add settings and theme options

makermelissa:
This PR adds a settings dialog that can be used to easily add new settings. The settings are persistent and are loaded at runtime. Currently I have added a theme setting.

This also includes changes to allow theming such as reorganizing the color-related CSS. I have created a light theme and if there's something I missed (e.g. light text on light background), we can always update it.

This adds features for the following issues:
#181, #354, #355


tannewt:
Please include a description of the changes in the PR itself. That way it'll be in the merge commit. Issue numbers are good to have but not enough on their own. Thanks!

PR #472, opened 2026-04-22
Fix infinite save retry loop and method check bugs

makermelissa-ai-assistant:
## Summary

Fixes #460 — Cannot Save code.py Edits on Adafruit Qualia ESP32-S3

### The Problem

When saving a file fails (e.g., due to a connection issue or board being busy), `saveFileContents()` catches the error and retries via `setTimeout(saveFileContents, 2000)` with **no retry limit**. This creates an infinite loop that:

- Retries the save every 2 seconds forever
- Causes the editor to appear to "reload" continuously (as reported in #460)
- Never shows the user an error message
- Continues even after disconnecting or switching to a different board

### Changes

**js/script.js:**
- Add a retry counter with a max of 3 attempts for `saveFileContents()`
- Show a clear error message to the user when retries are exhausted
- Reset retry counter on successful save
- Clear the retry timeout and reset counter on disconnect (prevents phantom retries after switching boards)
- Fix missing closing quote in save failure message template literal

**js/common/web-file-transfer.js:**
- Fix `_isMethodAllowed()`: `method.toUpperCase` was missing `()`, passing the function reference instead of calling it — so the method check always returned `false`
- Add missing `await` on `_isMethodAllowed()` call in `_fetch()` so the permission check result is actually evaluated (previously the unawaited Promise was always truthy, making the check dead code)

### Testing

- Build passes with `vite build`
- The Qualia-specific save failure likely has a firmware component too (display DMA contention during web server PUT), but this fix prevents the editor from looping forever regardless of the cause

### DIFF
--- a/js/common/web-file-transfer.js
+++ b/js/common/web-file-transfer.js
@@ -89,7 +89,7 @@ class FileTransferClient {
         };
 
         if (fetchOptions.method && fetchOptions.method.toUpperCase() != 'OPTIONS') {
-            if (!this._isMethodAllowed(fetchOptions.method)) {
+            if (!await this._isMethodAllowed(fetchOptions.method)) {
                 if (fetchOptions.method.toUpperCase() == "MOVE") {
                     // This should only happen if rename is used and the user doesn't have latest version
                     console.warn("Please upgrade to the latest version of CircuitPython. Allowing MOVE for now.");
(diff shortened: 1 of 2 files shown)


makermelissa (approved):
Looks good and I see there were some bugs you caught that had been previously missed.

PR #473, opened 2026-04-23
Pull in Beta Features

makermelissa:
This pulls in the latest Beta features into the main editor before the codebases get too far apart.


makermelissa-ai-assistant (commented):
Nice batch of improvements — settings/themes, the custom input dialog replacing `prompt()`, safe-mode detection, print support, and panel focus management all look solid. CI is green. A couple of things caught my eye:

---

**Bug — save retry calls the function immediately instead of deferring it**

In `js/script.js`, the save-retry change:

```js
currentTimeout = setTimeout(await saveFileContents(path), 2000);
```

This calls `saveFileContents(path)` right away (and `await`s it), then passes its resolved return value to `setTimeout` — so the 2-second delay never actually happens and retries fire immediately in a tight loop. The original code had the same conceptual issue (no `path` arg was forwarded), but the fix accidentally made it worse by adding `await`.

Should probably be:

```js
currentTimeout = setTimeout(() => saveFileContents(path), 2000);
```

---

**Minor — `_themes.scss` typo**

`--puctuation-color` is defined in both dark and light roots — looks like it should be `--punctuation-color`. It's used in the `.tok-punctuation` rule, so if someone tries to reference it by the correct spelling later it'll silently fall back.

---

**Nit — missing newlines at EOF**

A few files lost their trailing newline (`layout.js`, `_base.scss`, `dialogs.js`). Not a big deal, but some linters/diff tools get noisy about it.

---

Everything else looks great — the `isLocal()` logic fix, the comment fix for `isChromeOs()`, the theme CSS variable extraction, and the `InputModal` class are all clean. 👍

VaKo234 (commented):
MINOR

PR #474, opened 2026-04-23
Fix save retry loop, punctuation CSS typo, and missing EOF newlines

makermelissa-ai-assistant:
Fixes three issues spotted during review of #473:

### 1. Bug — Save retry fires immediately instead of after 2s delay

`setTimeout(await saveFileContents(path), 2000)` calls the function immediately (and `await`s it), then passes the resolved return value to `setTimeout`. This creates a tight retry loop instead of a 2-second backoff.

**Fix:** Wrap in an arrow function so `setTimeout` defers the call:
```js
setTimeout(() => saveFileContents(path), 2000)
```

### 2. Minor — CSS variable typo in `_themes.scss`

`--puctuation-color` → `--punctuation-color` (missing 'n') in both dark and light theme roots and the `.tok-punctuation` rule.

### 3. Nit — Missing trailing newlines

Added missing final newlines to `js/layout.js`, `sass/base/_base.scss`, and `js/common/dialogs.js`.


makermelissa (approved):
Thank you

PR #481, opened 2026-04-29
Add CircuitPython module highlighting overlay

### LINKED ISSUE #363: Add CircuitPython Types to Code Mirror
This would allow it to recognize additional variable/class types specific to CircuitPython and color code correctly.

makermelissa-ai-assistant:
Related to #363 (held open until beta features merge to main).

Targets the `beta` branch.

## Approach

CodeMirror 6 dropped the simple `extra_keywords` mechanism that CM5 had, so instead of forking `@codemirror/lang-python` (and tracking it forever) this layers a small `ViewPlugin` on top of the existing Python syntax tree.

The plugin walks the visible `syntaxTree`, finds `VariableName` nodes whose text matches a known CircuitPython core module or common Adafruit library, and tags them with class `tok-cp-module` so the theme can color them distinctly.

## Files

- **New:** `js/common/circuitpython_highlight.js` — the overlay extension.
- **Edited:** `js/script.js` — adds `circuitpythonHighlight` to `editorExtensions`.
- **Edited:** `sass/layout/_themes.scss` — adds `.tok-cp-module` rule with a magenta accent (`#FF79C6`).

## Module list

~125 names: all CircuitPython core modules (`board`, `digitalio`, `displayio`, `busio`, `wifi`, `synthio`, `keypad`, `pulseio`, `pwmio`, `usb_cdc`, `usb_hid`, `usb_midi`, etc.) plus the most common `adafruit_*` libraries plus `neopixel` and `simpleio`. The list lives in one `Set` at the top of the new file — easy to extend or trim.

## Why this approach

- ✅ No upstream maintenance burden — Python language stays stock, no fork of `@codemirror/lang-python`.
- ✅ Only marks identifier nodes (`VariableName`), so strings and comments are never miscolored.
- ✅ Cheap — decoration set is rebuilt only on doc/viewport/tree changes, and only visible ranges are walked.
- ✅ Zero new dependencies; uses `@codemirror/view` and `@codemirror/language` already in the tree.

## Trade-offs / future work

- Limited to bare identifier matching for now. To highlight CircuitPython-specific attributes (e.g. `board.D13`) we'd extend the same overlay to `PropertyName` nodes — easy follow-up.
- The module list is hand-maintained. Adding a name when a new core module ships is a one-line change.

## Test

- `npm run build` is clean.
- New class `tok-cp-module` ships in `dist/assets/index.css`.
- Verified the dev server starts and the editor renders Python files normally with the extension active.

### DIFF
--- a/js/common/circuitpython_highlight.js
+++ b/js/common/circuitpython_highlight.js
@@ -1,0 +1,200 @@
+// CircuitPython syntax highlighting overlay for CodeMirror 6.
+//
+// CodeMirror 6 dropped the simple `extra_keywords` mechanism that CM5 had,
+// so instead of forking @codemirror/lang-python we layer extra decorations
+// on top of the existing Python syntax tree. We walk the tree inside the
+// viewport, find identifier nodes whose text matches a CircuitPython name,
+// and tag them with a CSS class that the theme can style.
+
+import { ViewPlugin, Decoration } from "@codemirror/view";
+import { syntaxTree } from "@codemirror/language";
+
+// CircuitPython core/built-in modules. These are the names that appear in
+// `import foo` / `from foo import ...` inside CircuitPython code and are
+// the most consistent thing we can match without parsing semantics.
+//
+// Keep this list focused on modules that ship with CircuitPython (or are
+// extremely common Adafruit libraries). Anything we add here will be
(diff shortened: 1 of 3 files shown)


makermelissa (approved):
Tested and working

PR #488, opened 2026-05-06
Suggest firmware updates and auto-open device info post-connect

### LINKED ISSUE #357: Check if running latest firmware and suggest upgrade
Since we usually have 2 sets of firmware (stable and development), this would need to check based on certain conditions. Sometimes we only have a stable version too.

So it should check the latest stable and development releases. It should also check what the user is currently running and do something like this logic:

- If the user is running a development release (e.g. 10.0.0-alpha1):
  - If a higher stable release 10.0.0 is available, suggest that.
  - If a higher dev release (e.g. 10.0.0-beta1 is available) suggest that.
- If the user is running a stable release (e.g. 9.2.8):
  -  If a higher stable release (e.g. 9.2.9), suggest that AND
  - If a higher dev release (e.g. 10.0.0-beta1) also suggest that

This would need to be able to turn a version number into something numerical that could be compared.

makermelissa-ai-assistant:
Refs #357.

Adds a small `firmware-check` helper and surfaces "newer firmware available" suggestions in the **Discovery** and **Device Info** modals, plus auto-opens the Device Info dialog after a fresh connect across all three workflows (Web / USB / BLE) so the suggestion is visible without needing to click the Info button.

### What it does

- `js/common/firmware-check.js` parses CircuitPython version strings (e.g. `9.2.8`, `10.0.0-alpha.1`, `10.0.0-rc.0`, plus a tolerant `-dirty` suffix) into a comparable structure.
- It fetches the list of releases from `api.github.com/repos/adafruit/circuitpython/releases` (cached for the lifetime of the page) and picks the highest stable release and the highest dev pre-release.
- It then applies the rules from #357 to decide what to suggest:
  - If the user is on a **development** release:
    - Suggest the latest **stable** if it is newer than what they're on.
    - Suggest the latest **dev** release if it is newer than what they're on.
  - If the user is on a **stable** release:
    - Suggest a newer **stable** if any.
    - Suggest the latest **dev** release if any.
  - If only a stable release exists at all (no current dev), only the stable suggestion shows up.
- Both `DiscoveryModal` and `DeviceInfoModal` now render those suggestions under the device info table, linking to the board's download page on circuitpython.org so the user gets the build that actually exists for their board.

### Post-connect dialog

Previously only the Web workflow auto-opened a dialog after connecting; USB/BLE users had to click the Info button to see device info (and now the firmware suggestion). This PR moves the post-connect "show info" trigger into `loadEditor()` so all three workflows behave the same way:

- One-shot per connection session — silent reconnects do NOT re-pop the dialog.
- Reset on actual disconnect via `disconnectCallback`.
- Fire-and-forget so the rest of the post-connect flow (busy spinner, parameterized doc loading) isn't blocked while the user reads the info.
- Gated on `workflow.connectionStatus()` so it doesn't try to open before the workflow has fully connected.

`USBWorkflow.onConnected` was also reordered to run `super.onConnected(e)` before `loadEditor()` — previously `loadEditor` ran while `_connected` was still `CONNSTATE.partial`, which would have made the new gate skip the dialog on USB.

### Why GitHub releases?

`circuitpython.org` does publish per-board builds via `_data/files.json`, but that file isn't exposed as a public endpoint. The GitHub releases API is CORS-friendly, doesn't require auth for low-volume reads, and is canonical for stable vs prerelease tags. Per-board availability is left to the deep-link to the board page, which only lists firmware that was actually built for that board.

### UX

- The suggestion box is rendered into a container that is empty by default, so the dialog layout doesn't shift if the API hasn't responded yet — and `&:empty { display: none; }` hides the container entirely if there's nothing to suggest.
- API failures are non-fatal: a console warning is logged and the dialog renders unchanged.

### Notes for review

- Version comparison treats stable as ranked above any pre-release at the same X.Y.Z (e.g. `10.0.0` > `10.0.0-rc.0` > `10.0.0-beta.0` > `10.0.0-alpha.1`).
- Pre-release labels are ranked `alpha < beta < rc` and then by their numeric suffix.
- The repo doesn't have a JS test setup, so I verified the parser/comparator and suggestion logic with a quick Node import of the module locally; happy to paste those checks as a comment or to add a tiny browser sketch if useful.

### DIFF
--- a/index.html
+++ b/index.html
@@ -391,6 +391,7 @@
                 </tr>
             </tbody>
         </table>
+        <div id="firmware-update" class="firmware-update-suggestion-container"></div>
         <h3>More network devices<i class="refresh fa-solid fa-sync-alt" title="Refresh Device List"></i></h3>
         <div id="devices"></div>
         <div class="buttons centered">
@@ -432,6 +433,7 @@
                 </tr>
             </tbody>
         </table>
+        <div id="firmware-update" class="firmware-update-suggestion-container"></div>
         <div class="buttons centered">
             <button class="purple-button ok-button">Close</button>
         </div>
--- a/js/common/dialogs.js
+++ b/js/common/dialogs.js
@@ -1,4 +1,5 @@
 import {sleep, isIp, switchDevice} from './utilities.js';
+import {renderFirmwareSuggestions} from './firmware-check.js';
 import * as focusTrap from 'focus-trap';
 
 const SELECTOR_CLOSE_BUTTON = ".popup-modal__close";
@@ -357,6 +358,10 @@ class DiscoveryModal extends GenericModal {
         this._currentModal.querySelector("#mcuname").textContent = deviceInfo.mcu_name;
         this._currentModal.querySelector("#boardid").textContent = deviceInfo.board_id;
         this._currentModal.querySelector("#uid").textContent = deviceInfo.uid;
+        const updateContainer = this._currentModal.querySelector("#firmware-update");
+        if (updateContainer) {
+            renderFirmwareSuggestions(updateContainer, deviceInfo);
+        }
     }
 
     async _refreshDevices() {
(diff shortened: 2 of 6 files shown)


makermelissa-ai-assistant:
Pushed a follow-up commit that makes USB and BLE behave the same as Web: the Device Info dialog now auto-opens once after every fresh connect, so the firmware-update suggestion is visible without the user needing to click the Info button.

- Trigger moved from `checkConnected()` (Web-only branch) into `loadEditor()`, which all three workflows already call after a successful connect.
- A one-shot `shownDeviceInfoForCurrentSession` flag prevents silent reconnects from re-popping the dialog; the flag resets on disconnect.
- Dialog is opened fire-and-forget so the post-connect spinner / parameterized doc loading isn't blocked while the user reads the info.
- Removed the now-redundant explicit `showInfo()` calls in `checkConnected()` and the URL-backend bootstrap path.

makermelissa:
I tested this locally.

makermelissa (approved):
I'm going to merge before this becomes too stale.

Issue #361, opened 2025-09-05
Add more file color coding support


makermelissa:
Related comments from #20:
- Another idea I had was to support additional language color coding based on file extension that was opened. This would be fairly easy to add.

This could support stuff like XML, JSON, TXT or anything else that could run on the device and be edited.

makermelissa:
Merged in #515

PR #489, opened 2026-05-06
Add syntax highlighting for additional file types

makermelissa-ai-assistant:
Refs #361

Pick the CodeMirror 6 language extension based on the opened file's
extension, instead of always parsing the buffer as Python. Python
files keep the existing CircuitPython module highlight overlay, so
nothing changes for the common `code.py` / `boot.py` / `settings.toml`
flow.

### Behaviour

| Extension | Language plugin | Notes |
|---|---|---|
| `.py` | `lang-python` + CircuitPython overlay | unchanged |
| `.json` | `lang-json` | new |
| `.html`, `.htm` | `lang-html` | new |
| `.css` | `lang-css` | new |
| `.js` | `lang-javascript` | new |
| `.xml` | `lang-xml` | new |
| `.md` | `lang-markdown` | new |
| anything else (`.txt`, `.toml`, `.ini`, `.inf`, …) | none | plain text, still an improvement over running the Python parser on a non‑Python file |

New (untitled) documents still default to Python so the editor
behaves the same as before for the "create a fresh `code.py`" case.

### Implementation

* Added `@codemirror/lang-{json,html,css,javascript,xml,markdown}` to
  `package.json` (all CM6-compatible, same major version line as
  `lang-python`).
* In `js/script.js`:
  * Split the static `editorExtensions` list into a base list plus a
    per-file `languageExtensionsForPath(path)` helper.
  * `loadEditorContents(content, path)` now takes the path so the
    correct language plugin can be installed when `setState` runs.
  * `loadFileContents` forwards the path through to
    `loadEditorContents`.
  * Initial empty editor and "new document" still get Python +
    CircuitPython overlay (same as before).

### Testing

* `npm run build` succeeds.
* Manual verification (open .json / .html / .css / .js / .xml / .md /
  .py from a connected device) — each file gets the expected
  language coloring; CircuitPython module highlight only fires for
  `.py`.

### DIFF
--- a/js/script.js
+++ b/js/script.js
@@ -3,6 +3,12 @@ import { EditorView, keymap } from "@codemirror/view";
 import { EditorState } from "@codemirror/state";
 import { indentWithTab } from "@codemirror/commands"
 import { python } from "@codemirror/lang-python";
+import { json } from "@codemirror/lang-json";
+import { html } from "@codemirror/lang-html";
+import { css } from "@codemirror/lang-css";
+import { javascript } from "@codemirror/lang-javascript";
+import { xml } from "@codemirror/lang-xml";
+import { markdown } from "@codemirror/lang-markdown";
 import { syntaxHighlighting, indentUnit } from "@codemirror/language";
 import { classHighlighter } from "@lezer/highlight";
 import { circuitpythonHighlight } from "./common/circuitpython_highlight.js";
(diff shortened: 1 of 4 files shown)


makermelissa-ai-assistant:
Follow-up commit (`214b6ec`) addressing testing feedback:

1. **Save As now refreshes the language plugin.** The first commit only
   set the language on file load, so renaming `code.py` to `test.html`
   via Save As left the buffer parsed as Python. Moved the per-file
   language extension into a CodeMirror `Compartment` and dispatch a
   `reconfigure` effect from `saveFileContents` when the path changes,
   so the new extension's language plugin is applied immediately
   without rebuilding the editor state (cursor / undo history are
   preserved).

2. **Untitled docs default to Python** — already the case
   (`languageExtensionsForPath(null)` returns
   `[python(), circuitpythonHighlight]`); confirmed no change needed.

3. **Light-theme HTML readability.** The editor previously hardcoded
   `color: #ddd` as the default text color and used a palette tuned
   only for the dark background. On the light theme's `#f8f8f8`
   background, anything falling back to the default (notably HTML tag
   names — `tok-typeName`, which had no rule at all) rendered as
   nearly-invisible light-grey-on-white. Token colors are now CSS
   variables with light-theme overrides, and `.tok-typeName` has an
   explicit color so HTML/XML tag names actually highlight in both
   themes.

makermelissa-ai-assistant:
Found the actual bug — pushed `4179361`.

`workflow.saveFileAs()` mutates `this.currentFilename = path` **before** delegating down to `saveFile` → `saveFileContents`. My previous attempt compared `path !== workflow.currentFilename` inside `saveFileContents` to decide whether to reconfigure the language — but at that point they're already equal, so the dispatch never fired and the editor stayed on the old language.

Moved the language-switch trigger into `setFilename` (the single chokepoint all filename changes go through — Open File, New File, Save As, backend URL load) and added a module-local `editorLanguagePath` so we can compare against the editor's actually-installed language instead of `workflow.currentFilename`. Same-extension renames skip the reconfigure to avoid churning parser state.

Verified the dispatch logic against a Node sanity script (14 path/ext mappings + the Save As code.py→test.html scenario) before pushing — all green.

Worth re-testing the Save As code.py → test.html flow in light theme; the contrast fix from `214b6ec` only matters once the language actually swaps, which it now does.

makermelissa (approved):
Tested and it is working as expected.

PR #490, opened 2026-05-06
Fix Connect/Disconnect button behavior across all workflows

### LINKED ISSUE #373: Get Connect/Disconnect Button to work properly

makermelissa-ai-assistant:
Closes #373.

Fixes the Connect/Disconnect button getting stuck in a wrong state across **USB**, **Web**, and **BLE** workflows when the underlying connection is interrupted, plus the second-click UX.

### What was broken

From the issue (and tyeth's repro):

```
Uncaught (in promise) NetworkError: The device has been lost.
usb.js:277  Uncaught (in promise) NetworkError: The device has been lost.
script.js:592  Uncaught (in promise) NetworkError: The device has been lost.
```

When the board was unplugged (or the browser lost its USB grant, or the WiFi link dropped):

1. The editor didn't always notice it was disconnected → the **Disconnect** button stayed visible and did nothing.
2. Even when it did notice, clicking **Connect** afterwards re-prompted for a workflow instead of reconnecting to the same one.
3. Once the user picked a workflow, there was no way to back out to the chooser without closing/reopening the dialog.

#### 1. USB workflow — handle "device has been lost" cleanly (`js/workflows/usb.js`)

The unhandled rejections traced back to two places:

- `_readSerialLoop().catch(...)` called `onDisconnected()` with the default `reconnect=true`. That re-entered `connect() → connectToSerial() → _switchToDevice() → device.open()` on a dead device, which threw another `NetworkError` — the one that surfaced as `usb.js:277`.
- `serialTransmit()` had three sequential `await this.writer.ready / .write / .ready` calls; only two had `.catch()` handlers. The trailing `await this.writer.ready` was the source of the `script.js:592` rejection (terminal `onData` → `serialTransmit`).

Fixes:
- Read-loop catch now passes `reconnect=false`.
- `serialTransmit()` is wrapped in a single `try/catch`; any failure is treated as a disconnect (`onDisconnected(null, false)`), so we don't keep poking a dead writer.
- `onDisconnected()` is now **idempotent** (a small `_isDisconnecting` flag) — the WebSerial `disconnect` event and the read-loop rejection both fire on a yank, and previously both called us.
- If any of `reader.cancel()` / `writer.releaseLock()` / `serialDevice.close()` themselves throw `NetworkError` (because the device is gone), `safeReconnect` is forced to `false` so we don't fall back through the bad reconnect path.

#### 2. Web workflow — notice silent drops (`js/workflows/web.js`)

- `websocket.onerror` is now wired and fires `onDisconnected(event, false)`. Previously only `onclose` was handled, which doesn't reliably fire when the network goes away mid-session.
- `serialTransmit()` checks `websocket.readyState` and bails / disconnects if it's not `OPEN`, instead of throwing into a `try/catch` and silently logging.
- The PING_INTERVAL_MS keepalive watchdog (`_checkConnection`) was implemented but commented out — it's now re-enabled in `onConnected()`. A timed-out ping calls `onDisconnected(null, false)` so the UI flips back to **Connect**.

#### 3. Connect remembers the last workflow (`js/script.js`)

- A new `webeditor.lastBackend` localStorage entry is updated whenever `loadWorkflow()` actually loads a workflow.
- `checkConnected()` now prefers that backend and goes straight to its connect dialog instead of always opening the workflow chooser.
- A new `chooseAndShowConnect()` is exposed via the workflow `init` params (`chooseConnectionFunc`) so the connect dialogs can re-prompt for a workflow when the user wants to switch.

#### 4. "Choose a different workflow" link in each connect dialog

- `index.html` adds a `<p class="connect-back-row"><a class="connect-back">… Choose a different workflow</a></p>` to the BLE / Web / USB connect modals.
- `Workflow._wireBackToChooser(modal)` (new helper on the base class) attaches the click handler idempotently. Each subclass's `showConnect()` calls it after `getModal()`.
- A small block in `sass/layout/_layout.scss` styles the back link without shifting the existing modal layout.

### Testing notes

I'd suggest verifying:

- USB: connect → unplug board → Disconnect button flips to **Connect**, no `NetworkError` spam in console. Click **Connect** → goes straight back to the USB connect dialog.
- Web: connect to a board over WiFi → kill the network briefly → UI flips to **Connect** within ~5–10 s (ping watchdog).
- USB → Disconnect → click **Connect** → USB dialog re-opens; click "Choose a different workflow" → workflow chooser appears.
- BLE: a BLE-specific reconnect path was deliberately left alone (the existing `gattserverdisconnected` handler + the dialog's Reconnect button are how BLE was already behaving). Happy to follow up with a BLE-specific tightening if you'd like.

### Notes for review

- The `webeditor.lastBackend` value is only used as a hint — it never auto-attempts a hardware connection on its own; it just routes the user to the matching connect dialog. URL-supplied `?backend=…` still wins (handled in `getBackend()`).
- `_isDisconnecting` is reset in a `finally`, so a future failure inside the base `super.onDisconnected()` won't permanently jam the workflow.
- No new dependencies. `npm run build` passes locally.

### DIFF
--- a/index.html
+++ b/index.html
@@ -224,6 +224,7 @@
     <div class="popup-modal shadow connect-dialog closable" data-popup-modal="ble-connect">
         <i class="fa-solid fa-2x fa-xmark text-white bg-primary p-3 popup-modal__close"></i>
         <div id="ble-instructions" class="connection-layout content">
+            <p class="connect-back-row"><a href="#" class="connect-back"><i class="fa-solid fa-arrow-left"></i> Choose a different workflow</a></p>
             <p>If you are connecting to a new device, follow the steps below. Otherwise click the Reconnect button.</p>
             <p><button class="purple-button" id="bleReconnect">Reconnect</button></p>
             <section class="step">
@@ -268,6 +269,7 @@
     <div class="popup-modal shadow connect-dialog closable" data-popup-modal="web-connect">
         <i class="fa-solid fa-2x fa-xmark text-white bg-primary p-3 popup-modal__close"></i>
         <div id="web-instructions" class="connection-layout content">
+            <p class="connect-back-row"><a href="#" class="connect-back"><i class="fa-solid fa-arrow-left"></i> Choose a different workflow</a></p>
             <section class="step">
                 <div class="step-number"></div>
                 <div class="step-content">
@@ -309,6 +311,7 @@
     <div class="popup-modal shadow connect-dialog closable" data-popup-modal="usb-connect">
         <i class="fa-solid fa-2x fa-xmark text-white bg-primary p-3 popup-modal__close"></i>
         <div id="usb-instructions" class="connection-layout content">
+            <p class="connect-back-row"><a href="#" class="connect-back"><i class="fa-solid fa-arrow-left"></i> Choose a different workflow</a></p>
             <section class="step">
                 <div class="step-number"></div>
                 <div class="step-content">
--- a/js/script.js
+++ b/js/script.js
@@ -53,6 +53,34 @@ const messageDialog = new MessageModal("message");
 const connectionType = new ButtonValueDialog("connection-type");
 const settings = new Settings();
 
+// localStorage key used to remember the most recently chosen backend
+// ("web" | "ble" | "usb"). When the user clicks Connect after a
+// disconnect, we prefer the last backend over re-prompting for one.
+const LAST_BACKEND_KEY = "webeditor.lastBackend";
+
+function getLastBackend() {
+    try {
+        const name = window.localStorage.getItem(LAST_BACKEND_KEY);
+        if (name && isValidBackend(name)) {
+            return getBackendWorkflow(name);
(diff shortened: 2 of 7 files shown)


makermelissa (approved):
Tested and working perfectly.

PR #491, opened 2026-05-08
Document Linux USB mount-option workaround for OSError after Ctrl-D (#229)

### LINKED ISSUE #229: writing code.py causes: OSError: [Errno 5] Input/output error
### CircuitPython Version

```python
8.2.0
9.0.0
9.1.1
all of above tested on two RP2040 boards

also reproduced the problem on 9.1.1 on an ESP32-S3

did not reproduce on a Trinket M0
```

### Workflow(s)

USB in code-beta.circuitpython.org on 2024-08-07, 2019 ET

### Browser(s)

```python
Version 127.0.6533.99 (Official Build) (64-bit) on Ubuntu 24.04
```

### Console Log

```python
Load different workflow
index.js:91 Requesting any serial device...
index.js:91 > Requested  SerialPort {onconnect: null, ondisconnect: null, readable: null, writable: null}
index.js:91 switch to SerialPort {onconnect: null, ondisconnect: null, readable: null, writable: null}
index.js:93 Read Loop Init
index.js:93 Read Loop Started
index.js:91 Getting Device UID...
index.js:93 Device UID: F6254209F45633
index.js:91 Successfully loaded directory: CIRCUITPY
index.js:91 New folder name: CIRCUITPY
index.js:91 SerialPort {onconnect: null, ondisconnect: null, readable: ReadableStream, writable: WritableStream}
index.js:89 Connected!
index.js:89 Path: /code.py
index.js:93 Current File Changed to: /code.py

[no output when writing code.py]
```

### Steps

1. Shift reload https://code-beta.circuitpython.org
2. select USB and choose serial and CIRCUITPY for an RP2040 board (I've tried two). The board has previously had `storage.erase_filesystem()` done, so CIRCUITPY is in good shape.
3. open Serial window
4. edit `code.py` and another another `print()` statement. Note that saving code.py does NOT cause an auto-reload
5. Type ctrl-D in serial window. I get:
```
>>> 
soft reboot

Auto-reload is on. Simply save files over USB to run them or enter REPL to disable.
code.py output:
OSError: [Errno 5] Input/output error

Code done running.

Press any key to enter the REPL. Use CTRL-D to reload.
```
The `OSError` is new. It did not show up if I did ctrl-D before editing the file.

### Description

_No response_

### Additional information

**This does _not_ seem to reflect a corrupted CIRCUITPY.** Instead, it is some other I/O error, maybe due to something being done via serial. If I disconnect from web-editor and just connect via a terminal program, without resetting, and do a ctrl-D, I do not see the `OSError`.

makermelissa-ai-assistant:
Fixes #229.

## What

On Linux, the default `vfat` mount options for the CIRCUITPY USB MSC drive defer writes via the kernel page cache. When CircuitPython is soft-rebooted (<kbd>Ctrl</kbd>+<kbd>D</kbd>) immediately after a save from the web editor, the device may see an inconsistent filesystem and report:

```
OSError: [Errno 5] Input/output error
```

This is a host-OS behavior, not a web-editor or firmware bug. macOS and Windows are unaffected. Network and Bluetooth workflows are also unaffected — only the USB workflow on Linux.

## Why this approach

I spent a fair bit of time exploring code-side workarounds (`createWritable({mode: "exclusive"})`, post-write file re-open, REPL nudges, warm-up flushes, etc.). None of them were reliably fixing the bug — the writes really are sitting in the host kernel page cache for up to ~30 s, and the browser File System Access API does not expose any way to force `fsync()`. Verified empirically: remounting CIRCUITPY with `-o sync` makes the bug disappear in 3/3 trials. Restoring the default mount brings it back.

So instead of a fragile JS workaround, this PR documents the host-OS fix and surfaces it at the right moment — when a Linux user is opening the USB connect dialog.

## Changes

- **`js/common/utilities.js`**: add `isLinux()` helper. Uses `navigator.userAgentData.platform` first (modern Chromium) and falls back to `userAgent`, explicitly excluding Chrome OS and Android to avoid false positives.
- **`index.html`**: add a collapsed `<details>` notice in the USB connect dialog's "Select USB Host Folder" step. Hidden by default; shown only when `isLinux()` is true.
- **`js/workflows/usb.js`**: in `showConnect()`, query the notice element and toggle its `hidden` attribute based on `isLinux()`.
- **`sass/layout/_layout.scss`**: styling for the notice — soft-yellow info box with code-block formatting and a custom disclosure triangle.
- **`docs/LINUX_USB_MOUNT.md`**: full guidance document covering:
  - What's happening (kernel page cache + vfat default async)
  - Quick fix (`udisksctl unmount` + `udisksctl mount -o sync`)
  - Permanent fix (`udev` rule using `ENV{ID_FS_LABEL}=="CIRCUITPY"` and `UDISKS_MOUNT_OPTIONS+="sync"`)
  - Trade-offs

## What this PR does **not** change

- **No code changes to `js/common/fsapi-file-transfer.js`** — kept pristine.
- **No changes to the read-only fallback path** — issue #327 behavior preserved.
- **No changes to firmware, REPL behavior, or the web/BLE workflows.**

## Verification

- Linux (Pi 4B) on a CircuitPython board (RP2040): default mount → OSError on save+Ctrl-D; `mount -o sync` → no OSError in 3/3 trials.
- Build clean (`npm run build`).
- Notice renders correctly when `isLinux()` is true; hidden when false.
- Verified the dialog is no longer scroll-clipped on a 720p viewport with the collapsed disclosure.

### DIFF
--- a/docs/LINUX_USB_MOUNT.md
+++ b/docs/LINUX_USB_MOUNT.md
@@ -1,0 +1,113 @@
+# Linux USB Mount Notes (CIRCUITPY)
+
+If you use the CircuitPython Web Editor on Linux with the **USB workflow**
+and you see
+
+```
+OSError: [Errno 5] Input/output error
+```
+
+in the serial terminal after pressing <kbd>Ctrl</kbd>+<kbd>D</kbd> following
+a save, this page is for you. The cause is in the host operating system,
+not the editor and not your CircuitPython device.
+
+## What is happening
+
+When the browser writes to `code.py` (or any other file on the CIRCUITPY
+drive), it goes through the operating system's filesystem layer. On Linux,
+the default behavior of the `vfat` filesystem on a USB Mass Storage Class
+(MSC) device is to **buffer those writes in the kernel's page cache** and
+only push them to the device some time later (the kernel default is up to
+~30 seconds).
+
+When you then press <kbd>Ctrl</kbd>+<kbd>D</kbd>, CircuitPython on the
+device tries to import `code.py` immediately. If the host hasn't yet
+flushed its writes, the device sees an inconsistent filesystem and
+returns `OSError: [Errno 5] Input/output error`.
+
+This only affects **Linux** with the **USB MSC workflow**. macOS and
+Windows do not have this issue. Network and Bluetooth workflows are also
+unaffected.
+
+## Quick fix (current session)
+
+Remount the CIRCUITPY drive with the `sync` mount option so writes are
+sent to the device synchronously:
+
+```sh
+udisksctl unmount -b /dev/sdX1
+udisksctl mount -b /dev/sdX1 -o sync
+```
+
+Replace `/dev/sdX1` with the actual block device. Find it with:
+
+```sh
+lsblk
+# or
+mount | grep CIRCUITPY
+```
+
+After this, return to the editor, **reconnect** (the previous filesystem
+handle is invalidated by the remount), and resume editing.
+
+## Permanent fix: udev rule
+
+To make every CircuitPython device automount with `sync` going forward,
+add a `udev` rule.
+
+1. Find your board's USB Vendor ID (`idVendor`) and Product ID
+   (`idProduct`) using `lsusb`. CircuitPython boards often share VID
+   `239a` (Adafruit) but PIDs vary by board.
+
+   ```sh
+   lsusb
+   ```
+
+2. Create `/etc/udev/rules.d/99-circuitpython-sync.rules` with:
+
+   ```
+   # CircuitPython CIRCUITPY drive: mount synchronously to avoid
+   # OSError: [Errno 5] Input/output error from web-editor saves.
+   ENV{ID_FS_LABEL}=="CIRCUITPY", ENV{UDISKS_MOUNT_OPTIONS}+="sync"
+   ```
+
+   The label-based match catches any CircuitPython board, regardless of
+   VID/PID. If you'd rather scope it tighter, you can add additional
+   `ATTRS{idVendor}=="..."` clauses.
+
+3. Reload udev rules:
+
+   ```sh
+   sudo udevadm control --reload-rules
+   sudo udevadm trigger
+   ```
+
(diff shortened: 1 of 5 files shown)


makermelissa (approved):
Looks great

dhalbert:
@makermelissa I am not sure this is a great long-term solution, because it is a big deal for the user to add the udev rule, etc. On Chromebooks, it is impossible for the average user.

The problem is more that the "File System Access API" doesn't provide an fsync, and we should lobby for that.,

I spent years making sure various editors do fsync's after file write: https://learn.adafruit.com/welcome-to-circuitpython/recommended-editors, https://github.com/adafruit/Atom-fsync-on-save, etc.

Since there is a mechanism in the web editor now for doing file edits through the REPL for boards that don't present CIRCUITPY, would an alternative be to detect the editor is running on Linux, and revert to REPL-based file writes?

makermelissa:
The REPL based idea is worth exploring. It would require the user to eject the Circuit Python drive so it's not read only, but that's definitely a simpler solution. I'm out of hours this week, but I'll see if we can revert and take that approach instead.

makermelissa:
Another option is possibly to just ignore the error. The writes do happen immediately and I discovered that waiting 30 seconds before reloading avoided the error altogether.

makermelissa:
@dhalbert Ok, I tried going the REPL based route. However, the problem I'm running into is the device is set to read-only at boot. It doesn't seem like simply ejecting the drive makes it writable again. Do you know if there's an option to make it writable short of updating boot.py and restarting the device?

dhalbert:
No, we don't have a way to flip the write permissions without a remount, which would require a USB reset. Thanks for trying anyway.

I looked at https://developer.chrome.com/docs/capabilities/web-apis/file-system-access

> Caution: Changes are not written to disk until the stream is closed, either by calling close() or when the stream is automatically closed by the pipe.

I'm assuming you're closing the stream. If Chrome did an fsync after that, it would be helpful, but it does not.

So changing the mount options seems like the only reliable solution. The user could speed up the flushing just by typing `sync` in a terminal window, but there might still be a temporary I/O error.

PR #506, opened 2026-06-05
Broaden read-only filesystem messaging to cover non-host-mounted cases (refs #460)

makermelissa-ai-assistant:
## Summary

Refs #460 — the read-only filesystem dialog from PR #498 misdiagnoses
a real-world case the original reporter flagged: a board connected
to a USB **power-only** source (no host has CIRCUITPY mounted)
still hits the dialog and is told to disconnect a host that isn't
there.

This PR is a **messaging fix only** — the detection logic is
unchanged. It does **not** close #460; the underlying
"board-side filesystem stays locked while USB MSC is enabled even
when no host is using it" behavior is a firmware-side concern that
will be tracked separately.

### `js/script.js`

Connect-time read-only notice:

> ~~Filesystem is read-only — you can browse files, but saving will fail until USB Mass Storage is **released**.~~ [How to fix].
>
> **Filesystem is read-only — you can browse files, but saving will fail while USB Mass Storage is *enabled* on the board.** [How to fix].

Save-failure dialog:

> ~~The board's filesystem is locked, usually because CIRCUITPY is mounted on a computer over USB.~~
>
> **The board's filesystem is in read-only mode for web workflow. This happens whenever USB Mass Storage is enabled on the board (the default for most CircuitPython boards), whether or not a host computer is actively using it.**

Reordered the fix list (boot.py disable comes first now, because
the alternate "just unplug USB" remedy only helps the smaller
subset of users where a host actually has it mounted), and added
a heads-up paragraph:

> **Heads up:** disabling USB Mass Storage in `boot.py` means the
> CIRCUITPY drive won't appear on a host computer either — you'll
> edit through web workflow only. Many makers use a button or pin
> check in `boot.py` to choose between the two modes at startup.

That note is the missing piece for the dual-workflow case the
original reporter (#460) called out: they work on the same project
both at the bench (Mu over USB) and deployed (web workflow), and
the previous dialog's only fix recommendation forced them to pick one.

### `js/common/web-file-transfer.js`

The `_writeProtectedError().hint` string (used as the fallback
when the editor doesn't render the dialog itself) is updated to
match the new framing — points at USB MSC being enabled, doesn't
assert host involvement.

## Why not just close #460 with this?

The dialog is now accurate, but the **underlying board behavior**
(filesystem stays write-protected while USB MSC is enabled, even
on power-only connections) is still surprising and arguably wrong
on hardware that never sees a real host. Investigating whether
the CircuitPython firmware should release the lock when no host
has done a real `WRITE10` in some window, or expose a cooperative
"yield to web workflow" mechanism, is its own piece of work and
belongs in [adafruit/circuitpython](https://github.com/adafruit/circuitpython).
That follow-up will be tracked separately; #460 stays open as
the umbrella.

## Testing

- `node --check` on both edited files passes.
- `npm run build` (vite) passes clean.
- Diff is messaging-only; no control-flow changes, no new
  branches, no API surface change.

## Refs

- Refs #460
- Builds on #498 (which added the writeProtected detection in
  the first place — that logic is preserved verbatim).

### DIFF
--- a/js/common/web-file-transfer.js
+++ b/js/common/web-file-transfer.js
@@ -60,13 +60,20 @@ class FileTransferClient {
         const err = new ProtocolError("File System is Read Only.");
         err.status = 409;
         err.writeProtected = true;
-        err.hint = "The board's filesystem is currently locked, " +
-                   "usually because CIRCUITPY is mounted on a " +
-                   "computer over USB. Disconnect the USB cable, " +
-                   "or disable USB Mass Storage in boot.py, then " +
-                   "reset the board and try saving again. " +
-                   "(Ejecting the drive in your OS may not be " +
-                   "enough on its own.)";
+        // The board's filesystem is read-only to web workflow whenever
+        // USB Mass Storage support is enabled on the board, regardless
+        // of whether a host computer is actively using the drive. The
+        // common-case fix is to disable USB MSC in boot.py (which also
+        // hides the drive from the host); users who want both modes
+        // typically gate the disable on a button or pin check.
+        err.hint = "The board's filesystem is in read-only mode for " +
+                   "web workflow. This happens whenever USB Mass " +
+                   "Storage is enabled on the board (the default for " +
+                   "most CircuitPython boards), whether or not a host " +
+                   "computer has CIRCUITPY mounted. Disable USB Mass " +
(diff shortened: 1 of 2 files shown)


makermelissa (approved):
Looks good.

PR #511, opened 2026-06-08
Silently reconnect BLE after autoreload-induced disconnect

### LINKED ISSUE #377: BLE Disconnect on File Write Operations
### CircuitPython Version

```python
10.0.0-beta.3
```

### Workflow(s)

BLE

### Browser(s)

```python
Chrome (latest)
```

### Console Log

```python
[Violation] 'click' handler took 1097ms
index.js:21 Uncaught (in promise) Error: Modal has not been opened yet. No instance available
    at gi.getModal (index.js:21:14076)
    at DR._markStepCompleted (index.js:112:6919)
    at DR.connectionStep (index.js:112:6850)
    at DR.updateConnected (index.js:112:24682)
    at DR.onDisconnected (index.js:112:3749)
getModal @ index.js:21
_markStepCompleted @ index.js:112
connectionStep @ index.js:112
updateConnected @ index.js:112
onDisconnected @ index.js:112Understand this error
[Violation] 'click' handler took 4949ms
index.js:40 caught connection error NetworkError: Failed to execute 'getPrimaryService' on 'BluetoothRemoteGATTServer': GATT Server is disconnected. Cannot retrieve services. (Re)connect first with `device.gatt.connect`.
    at IE.checkConnection (index.js:40:4356)
    at IE.move (index.js:40:11590)
    at Ou._handleRenameButton (index.js:32:103194) NetworkError: Failed to execute 'getPrimaryService' on 'BluetoothRemoteGATTServer': GATT Server is disconnected. Cannot retrieve services. (Re)connect first with `device.gatt.connect`.
    at IE.checkConnection (https://code-beta.circuitpython.org/assets/js/index.js:40:4356)
    at IE.move (https://code-beta.circuitpython.org/assets/js/index.js:40:11590)
    at Ou._handleRenameButton (https://code-beta.circuitpython.org/assets/js/index.js:32:103194)
index.js:40 caught write error TypeError: Cannot read properties of null (reading 'writeValueWithoutResponse')
    at IE._write (index.js:40:4911)
    at IE.move (index.js:40:11812)
    at async DR.showBusy (index.js:112:4411)
    at async Ou._handleRenameButton (index.js:32:103156) TypeError: Cannot read properties of null (reading 'writeValueWithoutResponse')
    at IE._write (https://code-beta.circuitpython.org/assets/js/index.js:40:4911)
    at IE.move (https://code-beta.circuitpython.org/assets/js/index.js:40:11812)
    at async DR.showBusy (https://code-beta.circuitpython.org/assets/js/index.js:112:4411)
    at async Ou._handleRenameButton (https://code-beta.circuitpython.org/assets/js/index.js:32:103156)
index.js:40 caught write error TypeError: Cannot read properties of null (reading 'writeValueWithoutResponse')
    at IE._write (index.js:40:4911)
    at IE.move (index.js:40:11833)
    at async DR.showBusy (index.js:112:4411)
    at async Ou._handleRenameButton (index.js:32:103156) TypeError: Cannot read properties of null (reading 'writeValueWithoutResponse')
    at IE._write (https://code-beta.circuitpython.org/assets/js/index.js:40:4911)
    at IE.move (https://code-beta.circuitpython.org/assets/js/index.js:40:11833)
    at async DR.showBusy (https://code-beta.circuitpython.org/assets/js/index.js:112:4411)
    at async Ou._handleRenameButton (https://code-beta.circuitpython.org/assets/js/index.js:32:103156)
index.js:40 caught write error TypeError: Cannot read properties of null (reading 'writeValueWithoutResponse')
    at IE._write (index.js:40:4911)
    at IE.move (index.js:40:11854)
    at async DR.showBusy (index.js:112:4411)
    at async Ou._handleRenameButton (index.js:32:103156) TypeError: Cannot read properties of null (reading 'writeValueWithoutResponse')
    at IE._write (https://code-beta.circuitpython.org/assets/js/index.js:40:4911)
    at IE.move (https://code-beta.circuitpython.org/assets/js/index.js:40:11854)
    at async DR.showBusy (https://code-beta.circuitpython.org/assets/js/index.js:112:4411)
    at async Ou._handleRenameButton (https://code-beta.circuitpython.org/assets/js/index.js:32:103156)
index.js:40 caught write error TypeError: Cannot read properties of null (reading 'writeValueWithoutResponse')
    at IE._write (index.js:40:4911)
    at IE.move (index.js:40:11903)
    at async DR.showBusy (index.js:112:4411)
    at async Ou._handleRenameButton (index.js:32:103156) TypeError: Cannot read properties of null (reading 'writeValueWithoutResponse')
    at IE._write (https://code-beta.circuitpython.org/assets/js/index.js:40:4911)
    at IE.move (https://code-beta.circuitpython.org/assets/js/index.js:40:11903)
    at async DR.showBusy (https://code-beta.circuitpython.org/assets/js/index.js:112:4411)
    at async Ou._handleRenameButton (https://code-beta.circuitpython.org/assets/js/index.js:32:103156)
```

### Steps

1. Make sure a CIRCUITPY drive is NOT showing
2. Connect to the board
3. Click Save As
4. Rename an existing file
5. Errors in console

### Description

I'm seeing at least 3 different kinds of errors and I'm not sure how interrelated they are:

 -There are click handler violations to begin with. This is caused by a click event showing the modal and not returning control in a reasonable amount of time. Not sure how to work around this.
- Getting `Failed to execute 'getPrimaryService' on 'BluetoothRemoteGATTServer': GATT Server is disconnected. Cannot retrieve services.` and not sure why since it seems to be connected just fine. Perhaps an issue with the simplified BLE Workflow.
- Getting `caught write error TypeError: Cannot read properties of null (reading 'writeValueWithoutResponse')` and I'm not sure what that means

### Additional information

_No response_

### LINKED ISSUE #509: Bluetooth workflow disconnects after every write/rename/delete operation (UX paper cut, fixable editor-side)
## Summary

When using the BLE workflow, the editor disconnects from the device every time a write-style file operation completes (save, rename, move, delete, mkdir). The user has to manually click reconnect to keep working. This behavior is present at least as far back as CircuitPython 8.2.7 and 9.1.0 (tested on a CLUE / nRF52840), so it isn't a regression — but it is a real UX paper cut and I think it's fixable purely on the editor side.

## Reproduction

1. Connect a CLUE (or any BLE-capable CP board) using the Bluetooth workflow at https://code.circuitpython.org.
2. Open the file dialog, rename any file (or save `code.py`, or create a folder).
3. The rename succeeds on-device, but the editor immediately drops to "Disconnected" and shows the reconnect button.
4. Have to click Reconnect to do anything else.

Tested on CLUE running CircuitPython 8.2.7, 9.1.0, 10.0.0, 10.0.1, 10.0.3, and 10.1.1 — same disconnect behavior on all of them.

## Why it disconnects (firmware side)

In `supervisor/shared/bluetooth/file_transfer.c`, after any mutating command completes (`WRITE`, `WRITE_DATA`, `DELETE`, `MKDIR`, `MOVE`) the firmware calls `autoreload_trigger()` to reload the running Python program. This is intentional and matches USB MSC behavior ("save a file → `code.py` reruns"). The problem specifically over BLE is that the BLE-FT GATT server lives inside the VM that just got reloaded, so the GATT connection drops as a side effect of the reload.

This has been the design since BLE workflow shipped, which is why it's present in 8.2.7 going forward.

Re-architecting this on the firmware side (moving BLE-FT into supervisor-only code that survives reload) is a fairly invasive change, and would need testing across every BLE-capable port. I don't think we should chase that here — it's much cleaner to handle this on the editor side.

## Why the editor's existing reconnect doesn't kick in

`Workflow.onDisconnected(e, reconnect = true)` in `js/workflows/workflow.js` is supposed to attempt `this.connect()` when the disconnect was unexpected, and the `gattserverdisconnected` listener attached in `ble.js` invokes it with `reconnect` defaulting to `true`. But in practice no reconnect happens after the autoreload-induced drop. A few likely reasons:

- The BLE `connect()` override in `js/workflows/ble.js`:
  ```js
  async connect() {
      let result;
      if (result = await super.connect() instanceof Error) {
          return result;
      }
      ...
  }
  ```
  Due to operator precedence, `result` ends up being `true` / `false` rather than the actual `super.connect()` return value, so the early-return-on-error logic isn't doing what it looks like it's doing.

- Even without that bug, `connectToBluetoothDevice` is called immediately. The board is still mid-reload at that point, so `gatt.connect()` will fail and we silently bail.

- `connectToSerial()` and `initFileClient()` are called unconditionally in `switchToDevice`, but `gattserverdisconnected` listeners aren't rewired with the same care on the reconnect path.

## Proposed fix (editor side)

Add an automatic reconnect path that:

1. **Detects "expected" autoreload-induced disconnects** — i.e. a disconnect that happened within ~5 seconds of completing a `writeFile` / `move` / `delete` / `makeDir` call (or unconditionally, if simpler).
2. **Waits a short grace period** (~1.5–2.5 s) so the firmware has time to finish the autoreload and bring the GATT service back up.
3. **Calls `connectToBluetoothDevice`** on the same device handle we already have permission for (no `requestDevice` prompt needed because we still have `bleDevice`).
4. **Retries** with backoff (e.g. 3 attempts at 1 s / 2 s / 4 s) before giving up and showing the manual reconnect UI.
5. **Suppresses the "Disconnected" toast / status flicker** for the duration of the silent reconnect attempt, so the UX feels like "saving…" rather than "disconnected".

While we're in there, fix the precedence bug in `ble.js#connect`:

```diff
-    if (result = await super.connect() instanceof Error) {
-        return result;
-    }
+    result = await super.connect();
+    if (result instanceof Error) {
+        return result;
+    }
```

## Why this is editor-side, not firmware-side

- The disconnect has existed since BLE workflow shipped; firmware fix is high risk, low reward.
- Web Bluetooth's `navigator.bluetooth.getDevices()` lets us reconnect to an already-permitted device without a user gesture, so we have everything we need to make this transparent.
- Same approach naturally handles other autoreload-induced drops (e.g. user's `code.py` calling `supervisor.reload()` while connected).

Happy to take a stab at the PR if there's interest.

## Environment

- Board: Adafruit CLUE (nRF52840)
- CP versions tested: 8.2.7, 9.1.0, 10.0.0, 10.0.1, 10.0.3, 10.1.1
- Browser: Chrome on macOS (same behavior expected on Edge / Chromium-based browsers)
- Workflow: Bluetooth (code.circuitpython.org)

makermelissa-ai-assistant:
Fixes #377. Closes #509 as duplicate.

## Problem

CircuitPython firmware autoreloads on every BLE-FT mutating
operation — `WRITE`, `WRITE_DATA`, `DELETE`, `MKDIR`, `MOVE` —
which tears down the GATT server. See
[`supervisor/shared/bluetooth/file_transfer.c`][1] (search for
`autoreload_trigger`).

Before this PR, every save / rename / delete / mkdir over BLE
required the user to click **Connect** again, even though the
`bleDevice` handle was still valid and the `FileTransferClient`
state was salvageable. The file dialog also dropped its handle and
operations failed silently if attempted during reconnect.

## Fix

1. **Silent reconnect path in `BLEWorkflow.connect()`**: when the
   disconnect happens within ~5s of a known mutating op, reconnect
   to the same `bleDevice` (already paired) without a user
   gesture. Backoff retries at 1500 / 2500 / 4000 ms.

2. **`_rebindAfterSilentReconnect()`**: re-fetch BLE-FT service and
   NUS serial characteristics. Crucially, **reuse the existing
   `FileTransferClient` instance** instead of calling
   `switchToDevice()` — the open `FileDialog` holds bound method
   references that would be orphaned by client replacement. The
   upstream `@adafruit/ble-file-transfer-js` library lazily re-
   fetches its internal `_transfer` characteristic via
   `checkConnection()` on the next op when `_transfer` is null
   (which `onDisconnected` already sets), so this works cleanly.

3. **Mutating-op wrapper in `ble-file-transfer.js`**: `writeFile`,
   `move`, `delete`, `makeDir` now:
   - Signal the workflow that a mutating op is in flight
   - `await super.<op>()` to fire the firmware command
   - `await workflow.awaitPostOpReconnect()` so the caller sees a
     live GATT connection when the promise resolves

4. **`awaitPostOpReconnect()`** polls `bleDevice.gatt.connected`
   for up to 4s post-op, then awaits `_silentReconnectPromise` if
   one is in flight. This handles the race between
   `super.<op>()` resolving and `gattserverdisconnected` firing.

5. **Post-reconnect settle delay (2s)**: even after GATT is up,
   the firmware VM is still booting (running `boot.py`, mounting
   CIRCUITPY, initializing BLE service). If the next mutating op
   fires too early, `filesystem_lock()` returns false and BLE-FT
   returns `STATUS_ERROR_READONLY`. The settle delay buys time for
   the filesystem mount/lock state to stabilize.

6. **Operator-precedence fix in `connect()`**: original code was

   ```js
   if (result = await super.connect() instanceof Error) { ... }
   ```

   `instanceof` binds tighter than `=`, so `result` was always
   a boolean. Parenthesized to:

   ```js
   if ((result = await super.connect()) instanceof Error) { ... }
   ```

## Testing

On CLUE (nRF52840) running CircuitPython 10.0.0 over BLE:

- ✅ Rename file → autoreload → silent reconnect → file dialog
  updates with new path → auto-save fires → second autoreload →
  silent reconnect → fully back online
- ✅ Mutating ops on a single editor session no longer require
  manual reconnect
- ✅ No regression on first-time connect or genuine disconnects
  (out-of-range / device powered off) — those still surface the
  normal manual-reconnect UI

## Out of scope (not addressed by this PR)

- **USB-MSC concurrency**: when CIRCUITPY is mounted on a host,
  writes are read-only by design. Already addressed in #460 / #506.
- **Firmware-side `filesystem_is_writable_by_python()` regression**
  from CP PR #10659: tracked in adafruit/circuitpython#10972. That
  affects the web workflow PUT path, not BLE-FT (which uses
  `filesystem_lock()`, a different gate).
- **Upstream BLE-FT `_resolve is not a function`**: noticed during
  testing that late notifications arriving during reconnect can
  crash `processListDirEntry` and friends. Pre-existing, not
  regressed by this PR; worth a follow-up upstream PR to guard
  all `process*` handlers.

[1]: https://github.com/adafruit/circuitpython/blob/main/supervisor/shared/bluetooth/file_transfer.c

### DIFF
--- a/js/common/ble-file-transfer.js
+++ b/js/common/ble-file-transfer.js
@@ -1,17 +1,9 @@
 import {FileTransferClient as BLEFileTransferClient} from '@adafruit/ble-file-transfer-js';
 //import {FileTransferClient as BLEFileTransferClient} from '../../../ble-file-transfer-js/adafruit-ble-file-transfer.js';
 
-// Wrapper for BLEFileTransferClient to add additional functionality.
-// Optionally accepts a workflow reference so that mutating ops can notify
-// the workflow about the impending firmware autoreload (see
-// circuitpython/web-editor#377).
-//
-// Mutating ops (write/move/delete/mkdir) trigger a CircuitPython VM
-// autoreload, which kills the GATT connection. We hold the op's promise
-// open until either (a) the connection is restored, or (b) the silent
-// reconnect window expires, so callers like FileDialog can chain
-// `await fileHelper.move(...); await this._openFolder();` without
-// blowing up on a torn-down GATT in the second await.
+// Wrapper that holds mutating-op promises open across the firmware
+// autoreload + silent reconnect, so callers see a live GATT on return.
+// See circuitpython/web-editor#377.
 class FileTransferClient extends BLEFileTransferClient {
     constructor(bleDevice, bufferSize, workflow = null) {
         super(bleDevice, bufferSize);
--- a/js/common/ble-file-transfer.js
+++ b/js/common/ble-file-transfer.js
@@ -1,10 +1,61 @@
 import {FileTransferClient as BLEFileTransferClient} from '@adafruit/ble-file-transfer-js';
 //import {FileTransferClient as BLEFileTransferClient} from '../../../ble-file-transfer-js/adafruit-ble-file-transfer.js';
 
-// Wrapper for BLEFileTransferClient to add additional functionality
+// Wrapper for BLEFileTransferClient to add additional functionality.
+// Optionally accepts a workflow reference so that mutating ops can notify
+// the workflow about the impending firmware autoreload (see
+// circuitpython/web-editor#377).
+//
+// Mutating ops (write/move/delete/mkdir) trigger a CircuitPython VM
+// autoreload, which kills the GATT connection. We hold the op's promise
+// open until either (a) the connection is restored, or (b) the silent
+// reconnect window expires, so callers like FileDialog can chain
+// `await fileHelper.move(...); await this._openFolder();` without
+// blowing up on a torn-down GATT in the second await.
 class FileTransferClient extends BLEFileTransferClient {
-    constructor(bleDevice, bufferSize) {
+    constructor(bleDevice, bufferSize, workflow = null) {
         super(bleDevice, bufferSize);
+        this._workflow = workflow;
+    }
+
+    _signalMutatingOp() {
+        if (this._workflow && typeof this._workflow.markMutatingOp === 'function') {
+            this._workflow.markMutatingOp();
+        }
+    }
+
+    async _awaitReconnectIfNeeded() {
+        if (this._workflow && typeof this._workflow.awaitPostOpReconnect === 'function') {
+            await this._workflow.awaitPostOpReconnect();
+        }
+    }
+
+    async writeFile(path, offset, contents, modificationTime, raw) {
+        this._signalMutatingOp();
+        const result = await super.writeFile(path, offset, contents, modificationTime, raw);
+        await this._awaitReconnectIfNeeded();
+        return result;
+    }
+
+    async move(oldPath, newPath) {
+        this._signalMutatingOp();
+        const result = await super.move(oldPath, newPath);
+        await this._awaitReconnectIfNeeded();
+        return result;
+    }
+
+    async delete(path) {
+        this._signalMutatingOp();
+        const result = await super.delete(path);
+        await this._awaitReconnectIfNeeded();
+        return result;
+    }
+
+    async makeDir(path, modificationTime) {
+        this._signalMutatingOp();
+        const result = await super.makeDir(path, modificationTime);
+        await this._awaitReconnectIfNeeded();
+        return result;
     }
 
     async readOnly() {
--- a/js/workflows/ble.js
+++ b/js/workflows/ble.js
@@ -3,7 +3,7 @@
  */
 
 import {FileTransferClient} from '../common/ble-file-transfer.js';
-import {CONNTYPE} from '../constants.js';
+import {CONNTYPE, CONNSTATE} from '../constants.js';
 import {Workflow} from './workflow.js';
 import {GenericModal, DeviceInfoModal} from '../common/dialogs.js';
 import {sleep} from '../common/utilities.js';
@@ -13,6 +13,28 @@ const bleNusCharRXUUID = 'adaf0002-4369-7263-7569-74507974686e';
 const bleNusCharTXUUID = 'adaf0003-4369-7263-7569-74507974686e';
 
 const BYTES_PER_WRITE = 20;
+
+// Tunables for silent auto-reconnect after firmware autoreload.
+// CircuitPython's BLE file transfer triggers an autoreload after every
+// mutating op (write/move/delete/mkdir), which tears down the GATT
+// connection. We wait briefly, then reconnect to the same already-paired
+// device without a user gesture. See circuitpython/web-editor#377.
+const RECONNECT_DELAYS_MS = [1500, 2500, 4000];
+const POST_OP_RECONNECT_WINDOW_MS = 8000;
+// How long awaitPostOpReconnect() waits for the disconnect to fire after
+// a mutating op completes. Observed up to ~2s on CLUE / nRF52840 between
+// MOVE_STATUS=OK arriving over BLE and the firmware's VM reset actually
+// tearing down the GATT server. If no disconnect happens by then, assume
+// the firmware didn't autoreload (e.g. some future CP version) and proceed.
+const POST_OP_DISCONNECT_GRACE_MS = 4000;
+// After GATT reconnects post-autoreload, the firmware VM is still
+// booting. boot.py runs, code.py may start, BLE service finishes
+// initializing, and the CIRCUITPY filesystem mount/lock settles. If we
+// let the next mutating op fire too early, the firmware returns
+// STATUS_ERROR_READONLY because filesystem_lock() can't be acquired
+// during this window. Empirically ~1.5s is enough on CLUE / nRF52840;
+// budget 2s for slower boards.
+const POST_RECONNECT_SETTLE_MS = 2000;
 
 let btnRequestBluetoothDevice, btnReconnect;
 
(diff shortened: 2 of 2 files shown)


makermelissa (approved):
Thank you

PR #512, opened 2026-06-08
Fix duplicate BLE connect/disconnect messages and 3rd-reconnect failure

### LINKED ISSUE #1: Adapt a design that offers a clean and fast introduction to Circuit Python.
I believe proper separation, clearly labeled functions with icons, and prioritized controls make a user interface easier and more fun to use. 

Circuit Python Online Concept

Here is a concept that I created to showcase these design principals.

### LINKED ISSUE #410: Repeated messages (duplicates) when using save+run or restart/reload with disconnect+reconnect usage.
### CircuitPython Version

```python
10.0.1
```

### Workflow(s)

BLE

### Browser(s)

```python
Chrome latest
```

### Console Log

```python

```

### Steps

Feather S3 reverse tft
Device seems to not close connections cleanly clicking disconnect and connect w/reconnect again - ends up with multiple connect and disconnect messages (repeat disconnect connect + use reconnect button for even more)

### Description

_No response_

### Additional information

_No response_

makermelissa-ai-assistant:
Closes #410.

## Symptoms

In the BLE workflow, repeated connect/disconnect cycles would:

- Log multiple `Connected to ...` and `disconnected` lines per single user action (1 → 2 → 3 → ... per cycle).
- After ~3 reconnect cycles, the **Reconnect** button would log `Attempting to connect to ...` for each paired device but never actually establish a connection, with no error.

Reproduces on Feather S3 Reverse TFT (reporter's hardware) and CLUE nRF52840 (my test rig) running CircuitPython 10.x. Chrome latest.

## Root causes

Five overlapping bugs in `js/workflows/ble.js`:

### 1. `.bind(this)` returning a fresh function each call

```js
this.bleDevice.removeEventListener("gattserverdisconnected", this.onDisconnected.bind(this));
this.bleDevice.addEventListener("gattserverdisconnected", this.onDisconnected.bind(this));
```

`removeEventListener` only matches by reference, so the remove was a no-op. Every `switchToDevice` / `_rebindAfterSilentReconnect` / `_loadDom` call added another listener on top of the old ones.

**Fix:** Cache each bound handler on the instance in the constructor and reuse the same reference for add/remove. Pattern already used in `usb.js`.

### 2. `disconnectButtonHandler` double-firing `onDisconnected`

```js
this.bleDevice.gatt.disconnect();   // fires gattserverdisconnected -> onDisconnected
await this.onDisconnected(e, false); // ALSO calls onDisconnected manually
```

Two trips through `onDisconnected` per disconnect click → two log lines, two UI updates, two `connect()` recursions.

**Fix:** Only call `onDisconnected` manually when the GATT layer was already torn down (`!gatt.connected`); otherwise let the GATT event drive cleanup.

### 3. `onAdvertisementReceived` not guarded against re-entry

BLE peripherals can deliver several ads in the same event-loop tick before `abortController.abort()` actually detaches the listener. The handler would run multiple times per device, calling `switchToDevice` and `onConnected` multiple times.

**Fix:** Closure-local `advHandled` boolean; bail on subsequent entries.

### 4. Stale `watchAdvertisements` exhausting Chrome's per-device quota

`reconnectButtonHandler` iterates ALL paired devices and starts `watchAdvertisements` on each. Only one device responds; the others keep watching indefinitely. After several reconnect cycles, each device has multiple background watches, and Chrome silently stops delivering `advertisementreceived` events.

This is why the 3rd reconnect failed: the watches were active per Chrome's bookkeeping, but no events were being routed to our handlers.

**Fix:** Track every in-flight `AbortController` in `this._pendingAdvAborts` (Set). When any device wins, abort all of them (winner + losers). Also clear the Set on disconnect.

### 5. BLE-specific `onDisconnected` to detach lingering listeners

Even with fix #1, the device handle itself persists across reconnect cycles, and stale `gattserverdisconnected` listeners attached to old transitions could fire spuriously.

**Fix:** Add a BLE-specific `onDisconnected` override that detaches `gattserverdisconnected` from `bleDevice` and `characteristicvaluechanged` from `txCharacteristic` before delegating to `super.onDisconnected`.

## Test protocol

On a CLUE running CircuitPython 10.0.0, with 3 paired BLE devices visible to Chrome:

Before:
```
Attempting to connect to CIRCUITPY732a...
Connected to CIRCUITPY732a
disconnected
disconnected
Attempting to connect to CIRCUITPY732a...
Attempting to connect to CIRCUITPY732a...
Connected to CIRCUITPY732a
Connected to CIRCUITPY732a
disconnected
disconnected
disconnected
disconnected
... (3rd reconnect attempt logs "Attempting..." x3 but never connects)
```

After (verified by @makermelissa on Feather S3 Reverse TFT, 4+ cycles):
```
Attempting to connect to CIRCUITPY732a...
Attempting to connect to CIRCUITPY88bc...
Attempting to connect to CIRCUITPY4841...
Connected to CIRCUITPY732a
disconnected
Attempting to connect to CIRCUITPY732a...
Attempting to connect to CIRCUITPY88bc...
Attempting to connect to CIRCUITPY4841...
Connected to CIRCUITPY732a
disconnected
... (reliably reconnects every cycle, single log line per event)
```

(The N "Attempting" lines per cycle are expected: `reconnectButtonHandler` checks every paired device because we can't know in advance which one is on. Only the responsive one produces a `Connected`.)

## Out of scope

- The "fall-through to `connect()` after disconnect is a no-op when `bleDevice` is still set" path. The `reconnectButtonHandler` flow works because it doesn't depend on that path. A future PR could refactor to clear `bleDevice` on disconnect so auto-reconnect via `connect()` also works, but that interacts with the silent-reconnect path from #511 and deserves its own change.
- BLE-FT upstream library's unconditional `this._resolve(...)` calls during reconnect (separate issue noted in #511 work).

cc @makermelissa

### DIFF
--- a/js/workflows/ble.js
+++ b/js/workflows/ble.js
@@ -49,6 +49,19 @@ class BLEWorkflow extends Workflow {
         this._lastMutatingOpAt = 0;
         this._silentReconnectInFlight = false;
         this._silentReconnectPromise = null;
+
+        // Cached bound handlers so add/remove use the same reference.
+        // Without this, .bind() returns a fresh function every call and
+        // removeEventListener becomes a no-op, leaking listeners on every
+        // connect/reconnect cycle. See #410.
+        this._onRequestBluetoothDeviceClick = this.onRequestBluetoothDeviceButtonClick.bind(this);
+        this._onReconnectClick = this.reconnectButtonHandler.bind(this);
+        this._onSerialReceiveBound = this.onSerialReceive.bind(this);
+        this._onDisconnectedBound = this.onDisconnected.bind(this);
+
+        // Track in-flight watchAdvertisements abort controllers so we can
+        // cancel them when any device wins or when we tear down (#410).
+        this._pendingAdvAborts = new Set();
     }
 
     // Called by the FileTransferClient wrapper right before any mutating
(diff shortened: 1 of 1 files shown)


makermelissa (approved):
Looks good.

PR #528, opened 2026-07-15
Add Python formatter button

### LINKED ISSUE #362: Automatic Code Formatter
Related comments from #20:

- An automatic python code formatter (like pep8 or black)

This could be ust a button that when clicked will go through the code and format it. It would be great if we could use an external library and just run the code through there.

makermelissa-ai-assistant:
## Summary
- Add a Format button to desktop and mobile editor controls
- Format Python files in-browser using Astral's official Ruff WASM package
- Disable the formatter for non-Python files

## Testing
- npm run build
- Direct Ruff WASM smoke test formatting rough Python input

Refs #362


makermelissa (approved):
Looks good. Changes I requested through codex have been applied.

Issue #353, opened 2025-09-05
Adjust Font Size


makermelissa:
Add a way to adjust font size in the editor without having to zoom into everything

makermelissa:
#442 adds a settings panel. For this one, I mostly need to decide on the type of html control. I will most likely use a number input with minimums and maximums.

PR #531, opened 2026-07-17
Add editor font size setting

makermelissa-ai-assistant:
## Summary
- add a Font Size number input to the settings dialog
- clamp saved font sizes between 8px and 48px
- apply the saved font size to both CodeMirror and the Serial terminal on startup and after saving settings
- keep existing settings intact when the Settings dialog is canceled or closed

## Testing
- npm run build
- browser smoke test: changed Font Size from 16 to 22 and verified the editor computed font size changed to 22px
- verified the built bundle includes xterm fontSize updates and the settings-dialog cancel fallback

Refs #353

### DIFF
--- a/js/common/settings.js
+++ b/js/common/settings.js
@@ -29,7 +29,7 @@ class SettingsDialog extends GenericModal {
             contentDiv.appendChild(label);
 
             const control = await this._createControl(setting);
-            control.value = settings[setting.key];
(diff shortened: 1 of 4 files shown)


makermelissa (approved):
Tested locally, found a few bugs along the way, but it appears to be working well.

Issue #22, opened 2022-10-19
Display an icon to show what type of connection we have


makermelissa:
This will make it easier to switch between connections.

PR #536, opened 2026-07-22
Add connection type indicator

makermelissa-ai-assistant:
## Summary

Adds a visible connection type indicator to the Connect/Disconnect control so users can see whether the current or last selected workflow is WiFi, Bluetooth, or USB.

## Details

- Maps each workflow type to its existing Font Awesome icon and display label.
- Updates the header Connect/Disconnect buttons with the selected/remembered connection method, title text, and an accessible label.
- Keeps the mobile header compact by hiding the text label while preserving the icon.

Refs #22

## Validation

- npm run build
- git diff --check
- Browser check of the built app for initial state, WiFi selected state, and mobile header overflow


makermelissa (approved):
Looks good.

PR #537, opened 2026-07-23
Use serial file transfer on supported firmware

### LINKED ISSUE #399: Possibly Simplify USB Workflow
I would make the workflow much easier if we could only rely on the serial connection. This would make it so much easier to rename folders and use features like that. The main issue (and the reason it is the way it is now) is that I believe the file system is not writeable when the CIRCUITPY drive is present to avoid writing conflicts. We may be able to get around this by asking the user to eject the drive, and then writing
```
import storage
storage.disable_usb_drive()
```
to boot.py and and resetting. It could programmatically be removed from boot.py to re-enable the CIRCUITPY drive. Perhaps there is a way to just programmatically eject it, though I'm not seeing a way currently.

makermelissa-ai-assistant:
## Summary

- feature-probe `storage.unsafe_disable_usb_drive()` when the USB workflow
  connects to a read-only CIRCUITPY filesystem
- use REPL-based file transfer when the firmware can temporarily disable USB
  mass storage
- retain the existing File System Access API folder-picker workflow as the
  fallback for older or unsupported firmware
- restore the CIRCUITPY USB drive after a clean editor disconnect

## Why

New CircuitPython firmware support from
adafruit/circuitpython#11124 allows the editor to take CIRCUITPY offline after
USB has started. This makes the filesystem writable through the REPL and
removes the folder-operation limitations of the browser File System Access API.

The capability is detected at runtime rather than inferred from the firmware
version, so existing boards and firmware continue to use the current workflow.

Closes #399.

## Dependency

Uses `@adafruit/circuitpython-repl-js` 3.4.0, which includes the USB-drive
control helpers added by adafruit/circuitpython-repl-js#24.

## Validation

- `npm run build`
- `git diff --check`
- manually tested with the latest CircuitPython alpha firmware:
  CIRCUITPY disappeared during the serial editing session, REPL file operations
  worked, and the drive returned after disconnecting

### DIFF
--- a/js/workflows/usb.js
+++ b/js/workflows/usb.js
@@ -329,23 +329,8 @@ class USBWorkflow extends Workflow {
     }
 
     async _trySerialFileTransfer() {
-        const disabledMarker = "WEB_EDITOR_USB_DRIVE_DISABLED";
-        const unavailableMarker = "WEB_EDITOR_USB_DRIVE_UNAVAILABLE";
-        const code = `
-try:
-    import storage
-    if hasattr(storage, "unsafe_disable_usb_drive"):
-        storage.unsafe_disable_usb_drive()
-        print("${disabledMarker}")
-    else:
-        print("${unavailableMarker}")
-except Exception:
-    print("${unavailableMarker}")
-`;
-
         try {
-            const result = await this.showBusy(this.repl.runCode(code));
-            if (String(result || "").includes(disabledMarker)) {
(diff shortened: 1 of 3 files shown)


makermelissa:
While it tests functionally, we're going to move some of this code over to https://github.com/adafruit/circuitpython-repl-js

makermelissa (approved):
Tested and working.

PR #538, opened 2026-07-23
Move connection indicator to footer

makermelissa-ai-assistant:
## Summary

- move the connection type indicator out of the Connect/Disconnect button and into the footer
- restore the header buttons to plain Connect/Disconnect text
- keep the footer indicator icon-only with accessible title/ARIA text for connected or remembered workflows
- compact footer action buttons on small screens so the added icon does not cause horizontal overflow

Refs #22

## Validation

- npm run build
- git diff --check
- verified the built app on desktop and mobile-sized viewports with the WiFi workflow selected via `#backend=web`

### DIFF
--- a/index.html
+++ b/index.html
@@ -124,6 +124,7 @@
         <div id="footer-bar">
             <button id="btn-mode-editor" class="mode-button active">Editor</button>
             <button id="btn-mode-serial" class="mode-button">Serial</button>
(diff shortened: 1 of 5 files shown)


makermelissa:
I didn't like the look inside the button, so we moved it down to the footer where it's more subtle.

Issue #352, opened 2025-09-05
Multiple Tabs


makermelissa:
Related comments from #20:

- Adding some tabs and just displaying the active document in the serial terminal and editor
  - This would also be a bit tricky with juggling and transferring between URLs, but not impossible
- open mutliple buffers/files in separate tabs

Timeline8:
Adding my vote for this feature (if anyone is keeping score 👍). Mu currently allows multiple tabs which makes it easier to copy code from one program to another as well as having a "scratch" tab open for temporary parking of snippets of code.

PR #539, opened 2026-07-23
Add multiple editor tabs

makermelissa-ai-assistant:
## Summary
- add an editor tab strip so multiple files or scratch buffers can stay open at once
- preserve a separate CodeMirror state, active filename, and saved-change marker for each tab
- open files into new tabs, switch to already-open paths, and prompt before closing dirty tabs
- keep the existing serial/connection workflow shared across tabs

Refs #352

## Testing
- `npm run build`
- `git diff --check`
- browser smoke test at `http://127.0.0.1:4222/`: verified initial tab render, dirty tab marker after editing, no console errors, and no horizontal overflow at 390px mobile width


makermelissa:
I'm going to continue testing. I've found a number of small issue that have been resolved, but I'm not confident this is the final version until I test more thoroughly. So far it's looking great.

makermelissa (approved):
Ok, after some fairly thorough testing, I think this one is good to go.

PR #540, opened 2026-07-24
Fallback for serial USB drive helpers

makermelissa-ai-assistant:
## Summary
- fall back to inline CircuitPython snippets when `circuitpython-repl-js` does not expose `tryDisableUsbDrive()` / `enableUsbDrive()`
- preserves serial file transfer for supported beta firmware such as 10.3.0-alpha builds even when the bundled REPL helper API is unavailable

## Testing
- `npm run build`
- `git diff --check`

## Notes
Console error observed before this fix:

```text
Unable to disable the CIRCUITPY USB drive: TypeError: this.repl.tryDisableUsbDrive is not a function
```

### DIFF
--- a/js/workflows/usb.js
+++ b/js/workflows/usb.js
@@ -330,7 +330,25 @@ class USBWorkflow extends Workflow {
 
     async _trySerialFileTransfer() {
         try {
-            if (await this.showBusy(this.repl.tryDisableUsbDrive())) {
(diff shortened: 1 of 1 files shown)


makermelissa:
We're not going this route because it circumvents the advantage of putting the code in circuitpython-repl-js. It would require any code changes in 2 places.

Issue #12, opened 2021-12-19
Upload Files?


kyleplo:
Is there a way to upload files to my CircuitPython device using this web editor? If so, it should be easier to find. If not, it should be added.

tannewt:
You mean opening a local file and then sending it over?

makermelissa:
Not currently other than creating a new file and pasting in the text.

kyleplo:
Yes, I mean sending a local file. I wanted to do this for audio files, but this would also be useful for images and other binary files.

makermelissa:
I was thinking some file management utilities would be handy, but uploading would be as well and it could all be part of the same thing.

makermelissa:
@tannewt do you think https://github.com/adafruit/ble-file-transfer-js would handle binary files without modification?

tannewt:
> @tannewt do you think https://github.com/adafruit/ble-file-transfer-js would handle binary files without modification?

I'd like to think so but haven't tried it.

makermelissa:
Fair enough. I haven't tried either.

makermelissa:
I added code that works on Web Workflow and it _should_ work on BLE, though I can't seem to find my CLUE to test it. I'll stop by my office later and grab one to test with.

I _think_ it should work with how it is now, though there are extra parameters than JavaScript should just discard. If not, I have a fix for that in branch that also hasn't been tested.

makermelissa:
I tested and it caught a bug in the file transfer module. I created an issue here: https://github.com/adafruit/ble-file-transfer-js/issues/6. Also, it worked fine without accounting for the parameters.

makermelissa:
Fixed via https://github.com/adafruit/ble-file-transfer-js/commit/f85c66c2dec4b4019885eb3b7e2f4b8e3a4ab6a6 and 04dd38fa20344c85986b887272f8ff8133dff4f3.

PR #11, opened 2021-12-14
Correcting URL to live site in readme

ae5au:
The URL to the live site in README.md is linking to .com instead of .org.

PR #546, opened 2026-08-09
Bound the BLE connect and file-read waits so they report instead of hanging

dhalbert:

The BLE connect flow had several waits with no upper bound, so a failure that should have been reported instead left the editor sitting there indefinitely. This does not make the underlying failures go away — most are host-side and not ours to fix — it makes them report instead of hang.

## What was unbounded

**Waiting for an advertisement before connecting.** `connectToBluetoothDevice()` armed `watchAdvertisements()` and only called `gatt.connect()` from the `advertisementreceived` handler. That event is never delivered on Linux — Chromium's BlueZ backend gates it on an `EIR` D-Bus property that exists only in the ChromeOS fork, Web Bluetooth on Linux is officially ["partially implemented and not supported"](https://github.com/WebBluetoothCG/web-bluetooth/blob/main/implementation-status.md), and `watchAdvertisements()` is ["No longer pursuing"](https://chromestatus.com/feature/5180688812736512). So on Linux the connect was never attempted at all, which is the "device chooser works, nothing happens afterwards" symptom. There is now a bounded wait, and that timer is what drives the connect on Linux.

**`gatt.connect()` itself.** Chrome bounds this at roughly 41 s on Linux normally, but not while a `watchAdvertisements()` watch is armed — in that state the promise simply never settles. It is now raced against a timer and cancelled with `gatt.disconnect()`, which [Chrome has honoured as a cancel since M140](https://chromium-review.googlesource.com/c/chromium/src/+/6798921).

**The same call in `_attemptSilentReconnect()`.** Easy to miss and arguably worse, since CircuitPython autoreloads after every mutating file operation, so that reconnect ladder runs after every save. An unbounded connect there stalls the ladder, and the mutating op waits on it through `awaitPostOpReconnect()`, so a save spins with no way out.

**A file read on a dead link.** `readFile()` and `listDir()` could return a promise nobody can settle: the library installed its reject handler after writing the request, so a write that failed on a dead link had no reject to call, and `checkConnection()` swallowed its own failure rather than rethrowing, letting the read proceed anyway. Both are fixed upstream, in [adafruit/ble-file-transfer-js#13](https://github.com/adafruit/ble-file-transfer-js/pull/13) and [#14](https://github.com/adafruit/ble-file-transfer-js/pull/14), released as [1.1.0](https://github.com/adafruit/ble-file-transfer-js/releases/tag/1.1.0) — which also closes [#11](https://github.com/adafruit/ble-file-transfer-js/issues/11) and [#12](https://github.com/adafruit/ble-file-transfer-js/issues/12). This PR bumps the pin from `1.0.5` to `1.1.0` and takes the fix instead of working around it: a read on a dead link rejects because `checkConnection()` now rethrows, and a link that drops mid-read rejects because the response promise is installed before the request goes out. The device-info dialogs report the rejection instead of showing a blank dialog. An earlier version of this PR carried its own liveness guard in `js/common/ble-file-transfer.js` for the same symptom; with `1.1.0` pinned it is redundant and has been removed.

## Two smaller fixes

The advertisement watch is aborted **before** `gatt.connect()` rather than after. The reason is Chrome's per-device `watchAdvertisements()` quota (#410) — leaving losers armed piles up against it. An earlier version of this PR justified the same change by claiming that connecting while a BlueZ discovery session is active fails on Linux; that was withdrawn. The failures behind it were a faulty host controller, a MediaTek MT7920 that goes 0/40 while WiFi scans and 40/40 with the radio quiet, and the same board connects 20/20 on an Intel AX210 with a watch armed or not. The code comments record what survived.

The "Looking for …" status no longer names a device when several are being raced. Both reconnect paths loop over every permitted device, so the name shown was whichever was last in `getDevices()` order, not the one that would connect.

## Testing

Feather nRF52840 Express, Metro ESP32-S3 and Circuit Playground Bluefruit; CircuitPython 10.3.0-alpha.4; Chrome 151 on Linux (Intel AX210) and macOS.

Linux connect, serial and file write were exercised repeatedly over two days, including the bounded advertisement wait and its status message, which is what makes a Linux connect complete at all — the deployed editor still hangs on the connect dialog there. A failed connect now reports an actionable error and re-enables the button. The "Current Device Info" spinner no longer hangs when the link is down — it reports the failure. The bounded silent reconnect is the least exercised of these.

The `1.1.0` bump was re-tested on a Feather nRF52840 Express with the local read guard removed: reading device info, listing an empty directory ([ble-file-transfer-js#11](https://github.com/adafruit/ble-file-transfer-js/issues/11)), listing a 60-entry directory with varied name lengths ([#12](https://github.com/adafruit/ble-file-transfer-js/issues/12)), and saving `code.py` all work. With the board reset out from under a live connection, Current Device Info reports "Could not read device information. The connection to the device was lost." rather than spinning, as expected.

One known gap in the ordering change: `abort()` returns synchronously in page JS but the resulting `StopDiscovery` reaches BlueZ asynchronously in the browser process, so `gatt.connect()` may still fire while the kernel is scanning. Measurement says this does not matter — the kernel disables scanning about 1.5 ms before every create-connection regardless of what BlueZ believes.

## Context

Chrome registers no usable BlueZ pairing agent, so on Linux the board must be paired from the desktop's Bluetooth settings before the editor can use File Transfer. That is not something this PR can fix; there is detail in [adafruit/circuitpython#11178](https://github.com/adafruit/circuitpython/pull/11178).

Separate from this PR: #545 fixes Save As silently corrupting files.

### DIFF
--- a/js/workflows/ble.js
+++ b/js/workflows/ble.js
@@ -285,17 +285,21 @@ class BLEWorkflow extends Workflow {
             this._connectAttemptInFlight = true;
             clearTimeout(advTimer);
 
-            // This device won. Stop the OTHER devices' watches so they don't
-            // pile up Chrome's per-device watch quota. This device keeps its
-            // own watch until the connect settles: on Linux the kernel only
-            // takes the working connect path while a discovery session is
-            // active, and Chrome holds one for the lifetime of the watch.
-            this._abortAdvWatches(abortController);
+            // This device won. Stop every pending watch, this device's
+            // included, BEFORE connecting -- Chrome holds a BlueZ discovery
+            // session for as long as any watch is armed, and connecting while
+            // one is active is what fails on Linux. Measured by driving
+            // Device1.Connect() directly: discovery stopped 36/36, discovery
+            // active 18/52. Intermittent -- a host suspend/resume clears the
+            // failing state until the next boot, so it may not reproduce. In
+            // the failing case the HCI create-connection is identical to a
+            // working one and the controller simply transmits nothing until
+            // the attempt is cancelled ~20s later.
+            this._abortAdvWatches();
             try {
                 await this._connectToGattServer(device, reason);
             } finally {
                 this._connectAttemptInFlight = false;
-                this._abortAdvWatches();
             }
         };
 
--- a/js/workflows/ble.js
+++ b/js/workflows/ble.js
@@ -40,6 +40,11 @@ const ADVERTISEMENT_WAIT_MS = 2000;
 // measured from 0.5s (macOS, Windows) up to 26.6s (Linux), hence the generous
 // ceiling.
 const CONNECT_TIMEOUT_MS = 30000;
+// Per-attempt bound for the silent reconnect after a firmware autoreload.
+// Shorter than CONNECT_TIMEOUT_MS because this path runs once per entry in
+// RECONNECT_DELAYS_MS, and a reconnect that has not landed within ten seconds
+// has stopped being silent regardless of whether it eventually succeeds.
+const SILENT_RECONNECT_TIMEOUT_MS = 10000;
 
 let btnRequestBluetoothDevice, btnReconnect;
 
@@ -326,23 +331,33 @@ class BLEWorkflow extends Workflow {
     // Connect with a bound. gatt.connect() does not always reject on its own --
     // on Linux with a watch armed it never settles -- so race it against a timer
     // and cancel with gatt.disconnect(), which is the only way page JS can abort
-    // an in-flight connect.
-    async _connectToGattServer(device, reason) {
-        console.log(`Connecting to GATT Server from "${device.name}" (${reason})...`);
-        this.showConnectStatus("Connecting to " + device.name + "...");
-
+    // an in-flight connect. Chrome has honoured disconnect() as a cancel since
+    // M140; before that the attempt is orphaned rather than aborted, so treat a
+    // timeout as fatal rather than assuming the adapter is left clean.
+    async _connectWithTimeout(device, timeoutMs) {
         let connectTimer;
         try {
-            this.bleServer = await Promise.race([
+            return await Promise.race([
                 device.gatt.connect(),
                 new Promise((_, reject) => {
                     connectTimer = setTimeout(() => {
                         device.gatt.disconnect();
                         reject(new Error(
-                            `connect did not complete within ${CONNECT_TIMEOUT_MS / 1000}s`));
-                    }, CONNECT_TIMEOUT_MS);
+                            `connect did not complete within ${timeoutMs / 1000}s`));
+                    }, timeoutMs);
                 }),
             ]);
+        } finally {
+            clearTimeout(connectTimer);
+        }
+    }
+
+    async _connectToGattServer(device, reason) {
+        console.log(`Connecting to GATT Server from "${device.name}" (${reason})...`);
+        this.showConnectStatus("Connecting to " + device.name + "...");
+
+        try {
+            this.bleServer = await this._connectWithTimeout(device, CONNECT_TIMEOUT_MS);
         } catch (error) {
             console.log(error);
             // TODO(ericzundel): Add to suggestBLEConnectAction if we can determine the exception type
@@ -352,9 +367,6 @@ class BLEWorkflow extends Workflow {
             // Disable the reconnect button
             this.connectionStep(1);
             return;
-        }
-        finally {
-            clearTimeout(connectTimer);
         }
 
         if (this.bleServer && this.bleServer.connected) {
@@ -479,7 +491,11 @@ class BLEWorkflow extends Workflow {
                 await sleep(delay);
                 try {
                     console.log(`Silent reconnect: attempting after ${delay}ms…`);
-                    this.bleServer = await this.bleDevice.gatt.connect();
+                    // Bounded: an unbounded connect here stalls the whole
+                    // reconnect ladder, and every mutating op waits on it via
+                    // awaitPostOpReconnect(), so a save appears to hang.
+                    this.bleServer = await this._connectWithTimeout(
+                        this.bleDevice, SILENT_RECONNECT_TIMEOUT_MS);
                     if (this.bleServer && this.bleServer.connected) {
                         console.log('Silent reconnect: GATT reconnected, rebinding characteristics…');
                         await this._rebindAfterSilentReconnect();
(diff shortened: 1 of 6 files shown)


dhalbert:
Claude wrote this:

Noting a related gap found during the same testing, not addressed in this PR.

`FileTransferClient.readOnly()` in `js/common/ble-file-transfer.js` never detects anything — it returns before its own check:

```js
async readOnly() {
    let readonly = false;
    return false;
    // Check if the device is read only
    console.log("Checking if device is read only");
    // Attempt to write a 0-byte temp file and remove it
    const testPath = '/._ble_readonly_check';
    ...
}
```

Everything after the `return false` is unreachable. The other transports implement this for real — `fsapi-file-transfer.js` and `repl-file-transfer.js` both have working versions — so BLE is the exception rather than this being a convention.

The consequence is that when the board's filesystem is read-only to CircuitPython, which is the default whenever USB MSC is active, the editor reports it as writable and goes ahead with saves that cannot succeed. The device answers `STATUS_ERROR_READONLY`, the save retry loop runs its attempts, and the user eventually gets "Saving file ... failed after multiple attempts. Check your connection and try again." — which points at the connection, when the actual cause is that the drive is mounted over USB.

This is worth mentioning here because it made the failure hard to attribute during testing: repeated save failures over BLE looked like a bug in the write path, and it took a hex dump of the result to establish that nothing had been written at all and why.

The real check is not free — it writes and deletes a temp file on every connect — so short-circuiting it may well be deliberate. If so, the reporting is still worth improving: `STATUS_ERROR_READONLY` already comes back from the device on the first write attempt, so the message could name the real cause without any probing.

EDIT: Now filed as https://github.com/circuitpython/web-editor/issues/551, so it does not get lost when this PR merges. Re-verified 2026-08-18 against both main and this branch: readOnly() is unchanged, and STATUS_ERROR_READONLY is still the status the device returns on the first write.

Nothing in this comment is addressed by this PR.

makermelissa:
Regarding the readonly function, the return false may have been there temporarily to get around an issue it was having. It should only check for readonly on connection and then store that variable until disconnect so it isn't constantly writing a file.

js/workflows/ble.js
```diff
@@ -234,72 +262,165 @@ class BLEWorkflow extends Workflow {
...
+    // Connect with a bound. gatt.connect() does not always reject on its own --
+    // on Linux with a watch armed it never settles -- so race it against a timer
+    // and cancel with gatt.disconnect(), which is the only way page JS can abort
+    // an in-flight connect. Chrome has honoured disconnect() as a cancel since
+    // M140; before that the attempt is orphaned rather than aborted, so treat a
+    // timeout as fatal rather than assuming the adapter is left clean.
+    async _connectWithTimeout(device, timeoutMs) {
+        let connectTimer;
+        try {
+            return await Promise.race([
```
makermelissa, line 389: I think we could use the timeout function in https://github.com/circuitpython/web-editor/blob/main/js/common/utilities.js#L2 here, which basically already does this.

makermelissa (changes requested):
Just 1 small nitpick.

js/workflows/ble.js
dhalbert: Claude wrote this:

Switched to `timeout()`:

```js
async _connectWithTimeout(device, timeoutMs) {
    try {
        return await timeout(() => device.gatt.connect(), timeoutMs);
    } catch (error) {
        device.gatt.disconnect();
        throw error;
    }
}
```

`device.gatt.disconnect()` has to stay. `timeout()` abandons the promise, and an abandoned `gatt.connect()` keeps running in the browser process; `disconnect()` is the only way page JS can cancel it, and Chrome has honoured it as a cancel since M140. Calling it on any rejection rather than only on the timeout is what keeps this to a `try`/`catch` — `disconnect()` on a device that is not connected is a no-op.

Also needs `timeout` added to the `utilities.js` import. One small loss: a timeout now logs `Timed Out` rather than `connect did not complete within 30s`, which was console-only.

dhalbert: How does the above look to you? I can submit a commit for that?

makermelissa: Can we update timeout to work better? One of the reasons it's nice to have in one place is updating it improves it everywhere.

dhalbert: Claude wrote this:

Yes. Two things, for every caller: clear the timer when the callback wins, and pass the callback an `AbortSignal` that fires on timeout so it can stop what it started — `Promise.race()` abandons the loser, it doesn't cancel it. Today that leaves the BLE connect running in the browser process, and the connect wait in `web.js` polling every 100 ms for the life of the page.

```js
// Run the callback and if it doesn't complete in the given time, throw an error.
// The callback is given an AbortSignal that fires on timeout: a raced promise is
// abandoned, not stopped, so anything it started has to be cancelled explicitly.
async function timeout(callback, ms) {
    const controller = new AbortController();
    let timer;
    try {
        return await Promise.race([
            callback(controller.signal),
            new Promise((_, reject) => {
                timer = setTimeout(() => {
                    controller.abort();
                    reject(new Error("Timed Out"));
                }, ms);
            }),
        ]);
    } finally {
        clearTimeout(timer);
    }
}
```

Existing callers ignore the extra argument. The BLE side becomes:

```js
async _connectWithTimeout(device, timeoutMs) {
    return await timeout((signal) => {
        signal.addEventListener("abort", () => device.gatt.disconnect(), {once: true});
        return device.gatt.connect();
    }, timeoutMs);
}
```

Want the `web.js` loop fixed here too (`&& !signal.aborted`), or separately?

dhalbert:
I added another commit which updates to ble-file-transfer-js 1.1.0 and also removes some code that was compensating for the bugs that were fixed in 1.1.0. I retested on real hw to make sure.  -- @dhalbert

js/workflows/ble.js
makermelissa: We can do it in a separate pr.

makermelissa (approved):
I'm going to approve it at this point so that we can get something that works in place and then we can deal with minor things later.

dhalbert:
Thanks for merging. I am continuing to investigate BLE workflow issues, and I think there are some new and possibly superseding things I'll be PR'ing, some here, some in web-editor.

makermelissa:
I really appreciate a second set of eyes on this. It's been a complicated mess and I felt like I was hitting a wall on it.

dhalbert:
@makermelissa Sorry - in the future I am going to break down these kinds of conglomerate PR's into separate ones that are more easily reviewed.

makermelissa:
I just meant BLE was a mess and I was hitting a wall with trying to get it working in general.

PR #558, opened 2026-08-24
Update ble-file-transfer-js to 1.2.0

dhalbert:
Claude found the underlying bugs and wrote the library fixes.

Brings in [ble-file-transfer-js 1.2.0](https://github.com/adafruit/ble-file-transfer-js/releases/tag/1.2.0) ([adafruit/ble-file-transfer-js#15](https://github.com/adafruit/ble-file-transfer-js/pull/15)):

- Reads the transfer characteristic before subscribing, so pairing is triggered reliably. This makes espressif boards pair at all on firmware before [adafruit/circuitpython#11236](https://github.com/adafruit/circuitpython/pull/11236), and avoids doubled responses on Windows.
- Stops notifications before starting them, so reconnecting to a bonded board works. With 1.1.0, a Disconnect followed by a Connect hangs at "Current Device Info".
- Doubled or malformed responses are reported as errors instead of hanging a transfer forever.

#### Testing

Tested this library version on a Feather nRF52840 and a Metro ESP32-S3 from Chrome on Linux and on Windows 11, including first pairing and bonded reconnect on both.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -8,7 +8,7 @@
       "name": "web-editor",
       "version": "0.0.0",
       "dependencies": {
-        "@adafruit/ble-file-transfer-js": "adafruit/ble-file-transfer-js#1.1.0",
+        "@adafruit/ble-file-transfer-js": "adafruit/ble-file-transfer-js#1.2.0",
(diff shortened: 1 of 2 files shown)


makermelissa (approved):
Looks good.

PR #560, opened 2026-09-02
fix loading code file with '>>>' string

FoamyGuy:
This fixes an issue that causes opening a file to fail if the file contains ">>>" in a string. I believe that some part of the logic is getting that confused for the REPL prompt even if it is in a string literal. 

Fixed by transferring the contents of the file in base64 instead of raw.

Thested successfully on a ESP32-C6 with by opening a file that contains `print("hello >>> world")`

### DIFF
--- a/js/common/repl-file-transfer.js
+++ b/js/common/repl-file-transfer.js
@@ -31,14 +31,15 @@ class FileTransferClient {
 
     async readFile(path, raw = false) {
         await this._checkConnection();
-        let contents = await this._fileops.readFile(path, raw);
+        let contents = await this._fileops.readFile(path, true);
         if (contents === null) {
             return raw ? null : "";
         }
         if (raw) {
             return contents;
         }
-        return contents.replaceAll("\r\n", "\n");
+        let text = await contents.text();
+        return text.replaceAll("\r\n", "\n");
     }
 
     async writeFile(path, offset, contents, modificationTime, raw = false) {


makermelissa:
That makes sense because it goes through serial for that device, which uses circuitpython-repl-js. It needs to be more reliable because this is the favored way to open files.

makermelissa:
I don't think this is the correct fix because in certain situations, you want the text to be returned only and not the raw file. I need to look further, but the correct fix likely needs to be in https://github.com/adafruit/circuitpython-repl-js with mechanisms to only look for >>> in the right situations.

makermelissa-ai-assistant:
I traced this to the raw REPL response parser and opened adafruit/circuitpython-repl-js#25 with a fix and regression tests.

`_checkCodeRunning()` was calling `_detectCurrentMode()` before parsing each raw response. Since mode detection searches the accumulated serial buffer for `>>> ` and other prompt strings, text printed as raw stdout could be mistaken for a mode transition. The new PR keeps an active raw response under the documented `OK` / stdout / Ctrl-D / stderr / Ctrl-D / `>` framing and also handles `OK` split across serial chunks.

With that fixed in the library, this web-editor change should not need to route all text reads through the binary/base64 path and `Blob.text()`.

makermelissa:
Closing in favor of https://github.com/adafruit/circuitpython-repl-js/pull/25 and #561.

makermelissa:
The root-cause fix should be live in a couple minutes if you'd like to give it a test.

